PR #14, opened 2020-03-03
Add jupyterlab.

seanr15:


ybt195:
As long as lab 2.0.0 is tested, then this is good to go.

PR #24, opened 2020-04-15
Publish GCP images publicly

brain-good:


brain-good:
I tested this with a mini-build successfully -- cancelled the full build here to save some computation time.

PR #27, opened 2020-04-20
Remove tensorflow patches.

aaron276h:


brain-good:
You need to remove this as well: https://github.com/determined-ai/environments/blob/master/Makefile#L44.

Also can you bump this to 240 in a separate commit? https://github.com/determined-ai/environments/blob/master/Makefile#L20

aaron276h:
@brain-good done

PR #35, opened 2020-06-08
[DET-3184] Add a det-nobody user.

rb-determined-ai:
An alternate strategy I considered was using non-unique-uid users (the `--non-unique` option of `useradd`) to add an alias user to `nobody` with the same UID as the `nobody` user, but then when launching docker containers, to specify the user by name instead of by UID.  This is apparently a behavior that exists in Linux but from what I can tell it is discouraged as being non-standard and kinda fragile.

We could also have just set the `nobody` user's `HOME` to a real directory, but IMO it is best to leave the `nobody` user in its default state.

Users wanting to have the same features in custom images will have to duplicate these lines from our Dockerfiles, which is why I left comments explaining the situations where they are useful.


brain-good:
Just going to assume you tested it ;)

rb-determined-ai:
Yup, confirmed that building images with these lines on top of the default images and running them in determined with the `65533` uid and gid results in an image with a writable $HOME directory (tested with jupyter notebook shell)

PR #45, opened 2020-07-25
fix: Update version of Nvidia driver in cloud images.

neilconway:
Upgrade from 410.104 to 450.51.05. This apparently is necessary to get
T4 chips to work on G4 AWS instances. This driver *should* be compatible
with older versions of CUDA -- I tested with CUDA 10.0 only.

Along the way, switch from the "XFree86" package to the "Tesla" package.
It is a little unclear what the difference is, but the Nvidia download
site guides you to the latter when you're using enterprise GPUs, which
should always be the case in public cloud environments.

### DIFF
--- a/cloud/roles/nvidia-drivers/defaults/main.yml
+++ b/cloud/roles/nvidia-drivers/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
 # NVIDIA Driver Settings
-driver_version: "410.104"
+driver_version: "450.51.05"
 driver_script_name: "NVIDIA-Linux-x86_64-{{ driver_version }}.run"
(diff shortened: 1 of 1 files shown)


vishnu2kmohan:
> 410.104 to 450.51.05

@neilconway I've historically recommended & used the "[Latest Long Lived Branch Version](https://www.nvidia.com/en-us/drivers/unix/)" drivers (presently [450.57](https://www.nvidia.com/Download/driverResults.aspx/162107/en-us)), but you're right, NVIDIA themselves are directing us to [download](https://www.nvidia.com/Download/index.aspx) these "[Tesla](https://docs.nvidia.com/datacenter/tesla/index.html)" drivers.

> switch from the "XFree86" package to the "Tesla" package

Are there any licensing (or redistribution) restrictions that we may run afoul of, if we switch to these drivers? I assume not, but worth checking.

PR #47, opened 2020-08-27
Save parsable artifacts after publishing.

rb-determined-ai:
The artifacts will eventually be consumed by scripts in the main
Determined repository for automatically updating the image tags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ CPU_SUFFIX := -cpu
 CUDA_100_PREFIX := determinedai/environments:cuda-10.0-
 CUDA_101_PREFIX := determinedai/environments:cuda-10.1-
 GPU_SUFFIX := -gpu
+ARTIFACTS_DIR := /tmp/artifacts
 
 export CPU_TF1_ENVIRONMENT_NAME := $(CPU_PREFIX)pytorch-1.4-tf-1.15$(CPU_SUFFIX)
 export GPU_TF1_ENVIRONMENT_NAME := $(CUDA_100_PREFIX)pytorch-1.4-tf-1.15$(GPU_SUFFIX)
@@ -67,24 +68,23 @@ build-tf2-gpu:
 
 .PHONY: publish-tf1-cpu
 publish-tf1-cpu:
-	docker push $(CPU_TF1_ENVIRONMENT_NAME)-$(SHORT_GIT_HASH)
-	docker push $(CPU_TF1_ENVIRONMENT_NAME)-$(VERSION)
+	scripts/publish-docker.sh tf1-cpu $(CPU_TF1_ENVIRONMENT_NAME) $(SHORT_GIT_HASH) $(VERSION) $(ARTIFACTS_DIR)
 
 .PHONY: publish-tf2-cpu
 publish-tf2-cpu:
-	docker push $(CPU_TF2_ENVIRONMENT_NAME)-$(SHORT_GIT_HASH)
-	docker push $(CPU_TF2_ENVIRONMENT_NAME)-$(VERSION)
+	scripts/publish-docker.sh tf2-cpu $(CPU_TF2_ENVIRONMENT_NAME) $(SHORT_GIT_HASH) $(VERSION) $(ARTIFACTS_DIR)
 
 .PHONY: publish-tf1-gpu
 publish-tf1-gpu:
-	docker push $(GPU_TF1_ENVIRONMENT_NAME)-$(SHORT_GIT_HASH)
-	docker push $(GPU_TF1_ENVIRONMENT_NAME)-$(VERSION)
+	scripts/publish-docker.sh tf1-gpu $(GPU_TF1_ENVIRONMENT_NAME) $(SHORT_GIT_HASH) $(VERSION) $(ARTIFACTS_DIR)
 
 .PHONY: publish-tf2-gpu
 publish-tf2-gpu:
-	docker push $(GPU_TF2_ENVIRONMENT_NAME)-$(SHORT_GIT_HASH)
-	docker push $(GPU_TF2_ENVIRONMENT_NAME)-$(VERSION)
+	scripts/publish-docker.sh tf2-gpu $(GPU_TF2_ENVIRONMENT_NAME) $(SHORT_GIT_HASH) $(VERSION) $(ARTIFACTS_DIR)
 
 .PHONY: publish-cloud-images
 publish-cloud-images:
-	cd cloud && packer build -var "image_suffix=-$(SHORT_GIT_HASH)" environments-packer.json
+	mkdir -p $(ARTIFACTS_DIR)
+	cd cloud \
+		&& packer build -machine-readable -var "image_suffix=-$(SHORT_GIT_HASH)" environments-packer.json \
+		| tee $(ARTIFACTS_DIR)/packer-log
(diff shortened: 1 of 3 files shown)


scripts/push-and-log-docker.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -e
+set -x
+
+if [ ! "$#" -ne 4 ] ; then
+    echo "usage: $0 LOG_NAME BASE_TAG HASH VERSION" >&2
+    exit 1
+fi
+
+log_name="$4"
```
aaron276h, line 11: should this be `$1`?

Makefile
```diff
@@ -67,24 +67,23 @@ build-tf2-gpu:
 
 .PHONY: publish-tf1-cpu
 publish-tf1-cpu:
-	docker push $(CPU_TF1_ENVIRONMENT_NAME)-$(SHORT_GIT_HASH)
-	docker push $(CPU_TF1_ENVIRONMENT_NAME)-$(VERSION)
+	scripts/publish-docker.sh tf1-cpu $(CPU_TF1_ENVIRONMENT_NAME) $(SHORT_GIT_HASH) $(VERSION)
```
aaron276h, line 70: should all of these be: `push-and-log-docker` ?

scripts/push-and-log-docker.sh
```diff
@@ -0,0 +1,26 @@
...
+underscore_name="$(echo -n "$log_name" | tf - _)"
+
+docker push "$base_tag-$hash"
+docker push "$base_tag-$version"
+
+mkdir -p "/tmp/artifacts"
+
+log_file="/tmp/artifacts/publish-$log_name"
+
+echo "${underscore_name}_hashed: $base_tag-$hash" > "$log_file"
```
aaron276h, line 25: question: shouldn't all of these be `>>` since we are going to call this script multiple times?

scripts/push-and-log-docker.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+set -e
+set -x
+
+if [ ! "$#" -ne 4 ] ; then
+    echo "usage: $0 LOG_NAME BASE_TAG HASH VERSION" >&2
```
aaron276h, line 7: question: what does the `>&2` do here?

rb-determined-ai: echos to stderr

scripts/push-and-log-docker.sh
rb-determined-ai: fixed.

Makefile
rb-determined-ai: no, but I meant to rename that script.  Fixed.

scripts/push-and-log-docker.sh
rb-determined-ai: No, we're writing to a different file in each case.

scripts/publish-docker.sh
```diff
@@ -0,0 +1,26 @@
...
+underscore_name="$(echo -n "$log_name" | tf - _)"
+
+docker push "$base_tag-$hash"
+docker push "$base_tag-$version"
+
+mkdir -p "/tmp/artifacts"
+
+log_file="/tmp/artifacts/publish-$log_name"
+
+echo "${underscore_name}_hashed: $base_tag-$hash" > "$log_file"
```
dzhu, line 25: non-blocking: Making them all `>>` _would_ look a bit more consistent and avoid the risk of accidentally doing a bad copy in the future and ending up with two `>` lines. You can add a `: > "$log_file"` to explicitly truncate the file if there's any chance it might already exist. But this is pretty okay as it is.

You could also have `exec > "$log_file"` and then `echo` without redirection after that.

scripts/publish-docker.sh
```diff
@@ -0,0 +1,26 @@
...
+    echo "usage: $0 LOG_NAME BASE_TAG HASH VERSION" >&2
+    exit 1
+fi
+
+log_name="$1"
+base_tag="$2"
+hash="$3"
+version="$4"
+
+underscore_name="$(echo -n "$log_name" | tf - _)"
```
dzhu, line 16: blocking: `tr`, I think you mean (or `underscore_name="${log_name//-/_}"` if this were Bash).

Makefile
```diff
@@ -67,24 +67,23 @@ build-tf2-gpu:
...
 .PHONY: publish-tf2-gpu
 publish-tf2-gpu:
-	docker push $(GPU_TF2_ENVIRONMENT_NAME)-$(SHORT_GIT_HASH)
-	docker push $(GPU_TF2_ENVIRONMENT_NAME)-$(VERSION)
+	scripts/publish-docker.sh tf2-gpu $(GPU_TF2_ENVIRONMENT_NAME) $(SHORT_GIT_HASH) $(VERSION)
 
 .PHONY: publish-cloud-images
 publish-cloud-images:
-	cd cloud && packer build -var "image_suffix=-$(SHORT_GIT_HASH)" environments-packer.json
+	mkdir -p /tmp/artifacts
```
dzhu, line 86: non-blocking: Maybe make the artifact path a variable at the top of this file and an argument to `publish-docker.sh` for DRY's sake.

scripts/publish-docker.sh
rb-determined-ai: lmao the muscle memory

scripts/publish-docker.sh
rb-determined-ai: I just ran the `echo`s in a subshell to avoid this entirely.

Makefile
rb-determined-ai: sure

rb-determined-ai:
Tested `publish-docker.sh` manually on a throwaway image repository to confirm that it works ok.

scripts/publish-docker.sh
```diff
@@ -0,0 +1,28 @@
...
+docker push "$base_tag-$hash"
+docker push "$base_tag-$version"
+
+mkdir -p "$artifacts"
+
+log_file="$artifacts/publish-$log_name"
+(
+    echo "${underscore_name}_hashed: $base_tag-$hash"
+    echo "${underscore_name}_versioned: $base_tag-$version"
+) > "$log_file"
```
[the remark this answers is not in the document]
rb-determined-ai: thanks!

Makefile
```diff
@@ -67,24 +68,23 @@ build-tf2-gpu:
...
-	docker push $(GPU_TF2_ENVIRONMENT_NAME)-$(VERSION)
+	scripts/publish-docker.sh tf2-gpu $(GPU_TF2_ENVIRONMENT_NAME) $(SHORT_GIT_HASH) $(VERSION) $(ARTIFACTS_DIR)
 
 .PHONY: publish-cloud-images
 publish-cloud-images:
-	cd cloud && packer build -var "image_suffix=-$(SHORT_GIT_HASH)" environments-packer.json
+	mkdir -p $(ARTIFACTS_DIR)
+	cd cloud \
+		&& packer build -machine-readable -var "image_suffix=-$(SHORT_GIT_HASH)" environments-packer.json \
+		| tee /tmp/artifacts/packer-log
```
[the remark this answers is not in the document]
rb-determined-ai: thank you, fixed.

PR #48, opened 2020-09-11
Remove references to PEDL.

rb-determined-ai:
I think we will have to manually verify that the journald-to-cloudwatch stuff still works afterwards.


aaron276h (approved):

rb-determined-ai:
@aaron276h do you think we should bump to `0.6.0` for this change, due to the changing of some paths?  I'm concerned about landing as-is because users would immediately be building against it as `0.5.0`

rb-determined-ai:
we talked offline, we think it's fine.

PR #49, opened 2020-09-15
Add P3-instance regions.

rb-determined-ai:
Note that `eu-west-2` (London) does not have p2.8xlarge instance types (our default image type for `det-deploy`) which means it will require special handling.  However, the London region was specifically requested by an OSS user, so I think it's worth supporting.

All other regions added here support P3 instances and also the p2.8xlarge instances that we use by default.


seanr15 (approved):

aaron276h (approved):
🚀

PR #50, opened 2020-09-21
Build GCP imags with kdump

rb-determined-ai:
The purpose of this PR is not necessarily merge to master, it is to produce a usable GCP image with `kdump` enabled, which is needed to debug a customer bug.

If the `kdump` tooling proves useful, we can consider if enabling `kdump` by default is worth the runtime cost of reserving a section of RAM (the only real cost, AFAIK).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,11 @@ version: 2.1
 orbs:
   gcp-cli: circleci/gcp-cli@1.8.4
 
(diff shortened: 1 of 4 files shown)


rb-determined-ai:
Closing this PR but leaving the branch around.  It appears to be working for the customer.  We can rebase this branch in the future to get kdump images again based on some future version of master.

PR #61, opened 2020-12-15
chore: add daemontools package

dzhu:
The package provides `multilog`, a tool that we want to use in
Kubernetes to apply rotation to harness output log files.


stoksc (approved):
lgtm.

PR #62, opened 2021-01-20
chore: add CUDA 11 images and update to equivalent frameworks

mackrorysd:
This adds a new image based on CUDA 11 versions of all of NVIDIA's libraries, and the new framework releases that support it. It also bumps the other images to equivalent versions of the frameworks. There should have already been significant test coverage of these changes, as I've been running our GPU tests on the CUDA 11 images (see the -cuda11 tests here: https://app.circleci.com/pipelines/github/determined-ai/determined?branch=cuda11). All fixes for that run to be green have been merged into master already.


Dockerfile.gpu
```diff
@@ -82,8 +82,9 @@ RUN if [ "$TORCHVISION_PIP" ]; then ${CONDA_DIR}/bin/python3.6 -m pip install $T
 RUN if [ "$TENSORPACK_PIP" ]; then ${CONDA_DIR}/bin/python3.6 -m pip install $TENSORPACK_PIP; fi
 RUN ${CONDA_DIR}/bin/python3.6 -m pip install GPUtil
 
-ARG TORCH_CUDA_ARCH_LIST="3.7;6.0;6.1;6.2;7.0;7.5"
-RUN if [ "$TORCH_PIP" ]; then pip install --global-option="--cpp_ext" --global-option="--cuda_ext" git+https://github.com/determined-ai/apex.git@37cdaf4ad57ab4e7dd9ef13dbed7b29aa939d061; fi
+ARG TORCH_CUDA_ARCH_LIST
+ARG APEX_GIT
+RUN if [ "$APEX_GIT" ]; then pip install --global-option="--cpp_ext" --global-option="--cuda_ext" git+$APEX_GIT; fi
```
rb-determined-ai, line 87: Why parameterize this if we always set it for all gpu images?  (or do we?)

rb-determined-ai (approved):
Are we going to make the cuda 11 images the default?  Or what's the plan there?

Dockerfile.gpu
mackrorysd: Parameterizing this because the newest upstream version does not like being built in a CUDA 11 environment, but there are CUDA 11-specific fixes in it. So the CUDA 10 images will get our downstream (although maybe we should rebase it and only remove fixes we have cause to omit) and the CUDA 11 images will get the newer upstream.

mackrorysd:
> Are we going to make the cuda 11 images the default? Or what's the plan there?

Wasn't going to make it the default right away (although I think it works, even on CUDA 10 hardware). I'll update the default images to point to the latest frameworks, which enables a feature or two that require the newer frameworks. We'll publish the CUDA 11 image to support newer hardware.

One factor is that TF 1.15 is not supported by the community on CUDA 11. NVIDIA has a fork of it that they support on CUDA 11, but I ran into some issues we'll need to troubleshoot before we also ship a TF 1.15 image for CUDA 11.

Dockerfile.gpu
mackrorysd: Well, that's why I'm parameterizing APEX_GIT. TORCH_CUDA_ARCH_LIST has a similar issue where it can't contain the new compute architecture when you're building for CUDA 10, but it MUST be set when you're building for CUDA 11. Failure to do this results in it acting like you have no GPUs.

rb-determined-ai:
Cool.  We'll have to start dealing with `tools/scripts/bumpenvs.py` when we want to make it a default.

PR #64, opened 2021-02-16
chore: Publish images to AWS GovCloud [DET-4985]

mackrorysd:
Already added keys to do this to the determined-production context in CircleCI.

Haven't tested Packer locally - may still do that if it proves to be easy enough.

Note: using west because it has p2 instances (east doesn't), but pushing to east just in case it's ever needed there - AFAIK that's a trivial cost. Tried to go with a nearly identical image: recent build of Ubuntu Xenial (16.04) with the same specs and similar naming convention as what we use on non-GovCloud.


armandmcqueen (approved):

mackrorysd:
Did as much local testing as I could without duplicating credentials to complete my local environment. Found one bug: you can't have 2 "amazon-ebs" builders without giving at least one of them a unique name.

mackrorysd:
Merging - discussed changes offline with Armand, latest change was trivial and tested.

PR #65, opened 2021-02-17
chore: use Ansible from Ubuntu upstream

mackrorysd:
The previous installation procedure hits this error: "The repository 'http://ppa.launchpad.net/ansible/ansible/ubuntu focal Release' does not have a Release file." Indeed there's no distribution here for Focal (Ubuntu 20.04): http://ppa.launchpad.net/ansible/ansible/ubuntu/dists/.

Just using upstream Ansible for now until this can get resolved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,8 +31,11 @@ commands:
   install-ansible:
     steps:
       - run:
-          name: Install Ansible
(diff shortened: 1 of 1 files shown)


mackrorysd:
I asked on #ansible on FreeNode and was told this is because they're in the middle of a new release. They had to manually publish Ubuntu 20.04 packages since the last release, so they're working through some kinks for the first time now and expect the packages to be up, possibly by tomorrow.

PR #66, opened 2021-02-18
ci: push docker images to the NGC registry

shiyuann:
Push the docker images to the NGC registry.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,9 +5,17 @@ orbs:
   gcp-cli: circleci/gcp-cli@1.8.4
 
 commands:
-  docker-login:
+  login-docker:
+    parameters:
+      repository:
+        type: string
+        default: ""
+      username:
+        type: string
+      password:
+        type: string
     steps:
-      - run: echo "${DOCKER_PASS}" | docker login --username ${DOCKER_USER} --password-stdin
+      - run: echo "<<parameters.password>>" | docker login <<parameters.repository>> -u "<<parameters.username>>" --password-stdin
   activate-service-account:
     steps:
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -5,9 +5,17 @@ orbs:
   gcp-cli: circleci/gcp-cli@1.8.4
 
 commands:
-  docker-login:
+  login-docker:
```
rb-determined-ai, line 8: why reverse the name?  `docker-login` makes more sense since it's calling `docker login`

scripts/publish-docker.sh
```diff
@@ -13,12 +13,18 @@ base_tag="$2"
...
 
 underscore_name="$(echo -n "$log_name" | tr - _)"
 
 docker push "$base_tag-$hash"
 docker push "$base_tag-$version"
 
+docker tag "$base_tag-$hash" "$ngc_registry/$base_tag-$hash"
+docker tag "$base_tag-$version" "$ngc_registry/$base_tag-$version"
+docker push "$ngc_registry/$base_tag-$hash"
+docker push "$ngc_registry/$base_tag-$version"
```
rb-determined-ai, line 26: What's the plan for these images?  If you run our software in NGC, will you get them automatically?  Where will we encode that logic?

rb-determined-ai (dismissed):

.circleci/config.yml
shiyuann: hmm: it's just consistent with our naming in the determined repo.

scripts/publish-docker.sh
shiyuann: The plan is to just push the images to the NGC repo. Users can download images from the NGC registry. That's all I need to do. They need to manually specify the image name though.

.circleci/config.yml
rb-determined-ai: oh, ok.  works for me.

scripts/publish-docker.sh
rb-determined-ai: Ok.  Would it be possible to put it as the default task environment in the nvidia helm chart?  (Is there an ngc-related helm chart?)

shiyuann: Actually, I didn't make a separate helm chart for nvidia. It shares the same helm chart. They just need to override the registry path in the `values.yaml` for NGC images.

shiyuann:
@rb-determined-ai Fixed one bug and re-pushed it. So it needs your review for a second time.

PR #67, opened 2021-02-19
chore: also install Ansible Python package

mackrorysd:
If you don't install this, the workflow fails because it gets a non-zero status from "ansible-playbook --version". If you dig deeper, it's because some Python code in the CLI tries to import the Ansible Python module and it doesn't exist unless you've installed it. I've reported this on their IRC channel - seems to be the only point of contact I see on Launchpad.

I actually see this in Ubuntu 20.04 in general even when installing the stock / upstream Ansible deb package too. I'm still using the PPA we used to use because as of today, their Ubuntu 20.04 distribution is fixed with the new release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,8 +31,11 @@ commands:
   install-ansible:
     steps:
       - run:
-          name: Install Ansible
+          name: Install Ansible CLI
           command: sudo apt-add-repository --yes --update ppa:ansible/ansible && sudo apt install ansible
+      - run:
+          name: Install Ansible library
+          command: CRYPTOGRAPHY_DONT_BUILD_RUST=1 pip install ansible
 
   install-docker:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,11 +31,14 @@ commands:
   install-ansible:
     steps:
       - run:
+          name: Install Rust compiler (required for Python cryptography)
+          command: sudo apt install -y rustc
+      - run:
           name: Install Ansible CLI
           command: sudo apt-add-repository --yes --update ppa:ansible/ansible && sudo apt install ansible
       - run:
           name: Install Ansible library
-          command: CRYPTOGRAPHY_DONT_BUILD_RUST=1 pip install ansible
+          command: pip install ansible
 
   install-docker:
     steps:


.circleci/config.yml
```diff
@@ -31,8 +31,11 @@ commands:
   install-ansible:
     steps:
       - run:
-          name: Install Ansible
+          name: Install Ansible CLI
           command: sudo apt-add-repository --yes --update ppa:ansible/ansible && sudo apt install ansible
+      - run:
+          name: Install Ansible library
+          command: CRYPTOGRAPHY_DONT_BUILD_RUST=1 pip install ansible
```
rb-determined-ai, line 38: What's this CRYPTOGRAPH_DONT_BUILD_RUST about?

I'd prefer we just accept the cryptography guy actually has most of the world wrapped around his finger and we get used to installing rustc

rb-determined-ai (dismissed):

.circleci/config.yml
mackrorysd: From the blurb I got in the console about this, it sounded to me like it disabling the building of Rust bindings for Ansible and not that it would be missing secure functionality. Let me confirm with one other source at least...

mackrorysd: Oh no, seems it comes from here and will eventually be a hard fail: https://cryptography.io/en/latest/faq.html. *sigh* Having to use Linux packages, pip, and a Rust compiler just to use Ansible? That's gotten a little silly IMO.

rb-determined-ai: Oh, this was a huge flamewar in the last few weeks.  Get some popcorn, read through https://github.com/pyca/cryptography/issues/5771

rb-determined-ai: side note, on ubuntu, I think it's likely you could just `pip update pip` to some minimum version, and you wouldn't have to build from scratch.

mackrorysd: Thanks for the nightmare reading material. Installing Rust for now...

Makefile
```diff
@@ -45,10 +45,10 @@ build-tf2-cpu:
 .PHONY: build-tf1-gpu
 build-tf1-gpu:
 	docker build -f Dockerfile.gpu \
-		--build-arg BASE_IMAGE="nvidia/cuda:10.1-cudnn7-devel-ubuntu18.04" \
+		--build-arg BASE_IMAGE="nvidia/cuda:10.2-cudnn7-devel-ubuntu18.04" \
 		--build-arg TENSORFLOW_PIP="tensorflow-gpu==1.15.5" \
-		--build-arg TORCH_PIP="torch==1.7.1+cu101 -f https://download.pytorch.org/whl/torch_stable.html" \
-		--build-arg TORCHVISION_PIP="torchvision==0.8.2+cu101 -f https://download.pytorch.org/whl/torch_stable.html" \
+		--build-arg TORCH_PIP="torch==1.7.0 -f https://download.pytorch.org/whl/torch_stable.html" \
+		--build-arg TORCHVISION_PIP="torchvision==0.8.1 -f https://download.pytorch.org/whl/torch_stable.html" \
```
rb-determined-ai, line 51: is this meant to be in the same commit?

mackrorysd: Nope - I pushed too many commits out of my local fork. Fixed.

rb-determined-ai (approved):

.circleci/config.yml
mackrorysd: Having read more, I like the idea of upgrading pip. I'll put that on my list of things to start adding everywhere when convenient - for now I wanna get this unblocked.

PR #72, opened 2021-03-03
chore: fix Docker publishing code

mackrorysd:
* Use the complete DockerHub path when downloading for cloud images
* Only write to artifacts for the DockerHub downloads

### DIFF
--- a/scripts/publish-docker.sh
+++ b/scripts/publish-docker.sh
@@ -19,10 +19,12 @@ underscore_name="$(echo -n "$log_name" | tr - _)"
 docker push "$base_tag-$hash"
 docker push "$base_tag-$version"
 
-mkdir -p "$artifacts"
+if [ -z "$artifacts" ]; then
+    mkdir -p "$artifacts"
 
-log_file="$artifacts/publish-$log_name"
-(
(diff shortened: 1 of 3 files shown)


scripts/publish-docker.sh
```diff
@@ -19,10 +19,12 @@ underscore_name="$(echo -n "$log_name" | tr - _)"
 docker push "$base_tag-$hash"
 docker push "$base_tag-$version"
 
-mkdir -p "$artifacts"
+if [ -z "$artifacts" ]; then
```
rb-determined-ai, line 22: this if statement looks good but I think you'll have to change the `if [ "$#" -ne 5 ]` clause uptop

rb-determined-ai (commented):
Do you know of any way to make packer exit non-zero when not all builds are successful?  It seems insane to me that our circleci job did not show a failure.

That could be a problem for not-today though.

scripts/publish-docker.sh
mackrorysd: Oohh good catch

rb-determined-ai (approved):

PR #73, opened 2021-03-04
chore: Run Ansible playbook on both AWS builders

mackrorysd:
There are 2 Ansible playbook invocations, an AWS-specific one and a GCP-specific one. Right now neither runs for the GovCloud builder - so having the AWS one apply to both.


rb-determined-ai (approved):
lol

PR #74, opened 2021-03-05
chore: organize Docker files into scripts

mackrorysd:
The immediate goal is to make it easier to abstract the Python version out to a single variable, but we've talked for a while about wanting to make the Dockerfiles easier for end-users to modify, and part of that is cleaning it up and making it clearer what everything is for.


docker_files/install_pyflame.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
rb-determined-ai, line 1: Prolly just go ahead and delete this.  We do not use it anymore under any circumstances.

mackrorysd: Oh good to know. It's the only reason we need to expose the Python include directories, etc. when installing Python.

rb-determined-ai: great news!

Dockerfile.cpu
```diff
@@ -25,87 +24,48 @@ RUN apt-get update \
...
-	&& rm /etc/ssh/ssh_host_ed25519_key \
-	&& rm /etc/ssh/ssh_host_rsa_key
+		wget
 
-ARG CONDA_DIR="/opt/conda"
-ARG CONDA_INSTALLER="Miniconda3-4.7.10-Linux-x86_64.sh"
-ARG CONDA_MD5="1c945f2b3335c7b2b15130b1b2dc5cf4"
-ARG CONDA_URL="https://repo.continuum.io/miniconda"
-ENV CONDA_DIR="${CONDA_DIR}" PATH="${CONDA_DIR}/bin:${PATH}"
+COPY docker_files /tmp/det_docker_files
```
mackrorysd, line 29: I'm not a big fan of the name docker_files. dockerfile_helpers or something might be better?

Dockerfile.cpu
```diff
@@ -25,87 +24,44 @@ RUN apt-get update \
...
+RUN /tmp/det_docker_files/install_libnss_determined.sh
 
-# Add a plugin to the user system that lets us extend the users available in
-# the container at runtime. This is critical for supporting non-root shell,
-# which in turn is critical for non-root distributed training.
-COPY libnss_determined /tmp/libnss_determined
-RUN make -C /tmp/libnss_determined libnss_determined.so.2 install \
-  && rm -rf /tmp/libnss_determined \
-  && sed -E -i -e 's/^((passwd|shadow|group):.*)/\1 determined/' /etc/nsswitch.conf
+RUN /tmp/det_docker_files/cleanup.sh
```
rb-determined-ai, line 67: Honestly I'm not sure this one needs to get pulled out the way the others do.  There is the cache-size reduction which I think is pretty standard to run like `RUN apt instal ... && rm -rf /var/cache/apt` or something, and there are the deleting preconfigured ssh keys which is only required to pass the NGC security screen (even though those keys are totally unused in our system).

Dockerfile.cpu
rb-determined-ai: Why not just use the existing `files` directory?  Seems not terribly different from this new `det_docker_files` directory to me.

Dockerfile.cpu
```diff
@@ -25,87 +24,44 @@ RUN apt-get update \
 		openssh-client \
 		openssh-server \
 		pkg-config \
-		wget \
-	&& rm -rf /var/lib/apt/lists/* \
```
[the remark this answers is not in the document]
mackrorysd: Good point. Considered running a conda cleanup command at the end too and figured it made sense to have all the clean up stuff together. In the end that didn't make a noticeable difference (may still be worth doing), but yeah - I've put these back.

Dockerfile.cpu
mackrorysd: In fact the previous contents of files/ just got moved under here. I went with dockerfile_scripts. I feel like that's clearer - we had top-level directories named "files" and "scripts", and it's not terribly clear one is used in Dockerfiles and one is used for publishing the images... Thoughts?

rb-determined-ai: I don't care very much, tbh

rb-determined-ai (approved):

PR #77, opened 2021-03-25
chore: Moving back to Python 1.7 and Python 3.7

mackrorysd:
A couple of problems found in testing making us shift back to slightly older versions for now:
* Cloudpickle on Python 3.8 can fail to read checkpoints written with older versions, so using Python 3.7 across the board until that's addressed (or until we can do checkpoints in another format).
* Horovod isn't fully testing PyTorch 1.8 and there's a recent fix we need to pull into our fork from upstream. This should happen very soon. But for now, just being a little more conservative.
I've been running the post commits against these images in a private branch but will re-confirm.


hamidzr (approved):

mackrorysd:
Here's the testing: https://app.circleci.com/pipelines/github/determined-ai/determined/10880/workflows/79aeaee9-9af2-4c08-879c-7d6909af42f0. Regular GPU tests succeeded. CUDA-11 K8s tests failed on test_gang_scheduling, which is currently only in my local branch.

PR #79, opened 2021-04-07
Install a cpu wheel of TF2.4 in the cpu image

AdamHillier:
This should shave off about 250 MiB.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ build-tf1-cpu:
 build-tf2-cpu:
 	docker build -f Dockerfile.cpu \
 		--build-arg PYTHON_VERSION="3.7" \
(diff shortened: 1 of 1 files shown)


mackrorysd:
Thanks for the PR, Adam! I think this is good. I think you're our first community contributor to this repo - looks like we don't have all the same infra set up for contributing to this repo as we do on the main one. While I get that fixed, could I get you to submit the CLA? https://www.determined.ai/cla/ Sorry to be a nuisance for such a straightforward change!

AdamHillier:
Thanks! CLA now signed :)

mackrorysd (approved):
Thanks for doing that, Adam! We'll get this rolled in next time we update the images.

PR #82, opened 2021-04-08
chore: update agent images to Ubuntu 20.04 [DET-5238]

mackrorysd:
Ubuntu 16.04 is about to end-of-life, so we need to update. Moving to Ubuntu 20.04 so it's safer to run Ubuntu 20.04-based task environments too.


mackrorysd:
Images all come from here, by the way: https://cloud-images.ubuntu.com/locator/. I think there's a good chance this might break post-commit, but that's the fastest way to find the issues, so I'll just iterate from there.

stoksc (approved):
lgtm

mackrorysd:
Note that this was subsequently squashed into PR #82 in order to re-run CircleCI with a different hash but no code changes.

PR #83, opened 2021-04-08
chore: latest Ubuntu AMIs not yet available on GovCloud

mackrorysd:


stoksc (approved):

mackrorysd:
Note that this was subsequently squashed into PR #82 in order to re-run CircleCI with a different hash but no code changes.

PR #87, opened 2021-04-13
chore: switch to upstream hvd and nccl

stoksc:
This change updates dockerfile_scripts/install_nccl_and_horovod.sh to install Horovod and NCCL 

## Commentary

Gonna test images on latest master once the pr to allow 2 agents lands. Tested with one and it worked fine but that's useless lol.

### DIFF
--- a/dockerfile_scripts/install_nccl_and_horovod.sh
+++ b/dockerfile_scripts/install_nccl_and_horovod.sh
@@ -4,10 +4,5 @@ set -e
 
 DEBIAN_FRONTEND=noninteractive apt-get install -y libnccl-dev libnccl2 --no-install-recommends
 
-git clone https://github.com/determined-ai/nccl.git /tmp/det_nccl
-(cd /tmp/det_nccl && git checkout -q 8768cb3f881ce198825294b62a435e3f3a804baf)
-make -C /tmp/det_nccl -j 4
-ldconfig /usr/local/cuda/targets/x86_64-linux/lib/stubs 
-pip install git+https://github.com/determined-ai/horovod.git@4911d9358c4c8e6aa569a69c8692f85ab035419b
+pip install horovod
 ldconfig
-rm -rf /tmp/det_nccl


mackrorysd (approved):
Looks good to me. I recently tested an identical change. Not exhaustively, mind you, but enough to feel confident putting this in at the start of a release cycle.

However - I know one of the motivations of this for you is getting PyTorch 1.8. Those changes have still not been released. You'll be missing at least 2 other changes you need: https://github.com/horovod/horovod/compare/v0.21.3...master. I'm not sure we should be building from master from source. But if we do, we should figure out a way to record exactly what we built so it can be looked up when needed.

mackrorysd:
Also, this is probably worth bumping the VERSION file but we can do that in another release. I actually wonder if the VERSION should start tracking the current Determined version at the time, so we can roughly correlate them. Even though ideally, we very rarely break the contract of compatibility between the two.

stoksc:
> However - I know one of the motivations of this for you is getting PyTorch 1.8. Those changes have still not been released. You'll be missing at least 2 other changes you need: horovod/horovod@v0.21.3...master. I'm not sure we should be building from master from source. But if we do, we should figure out a way to record exactly what we built so it can be looked up when needed.

Yeah, I was thought about / checked this when I was wondering whether we should pin to a specific version for now and be more intentional about updating or not. Looks like they don't cut releases very frequently, but also I don't quite feel comfortable pinning to a git commit. I'll sleep on it.

> But if we do, we should figure out a way to record exactly what we built so it can be looked up when needed.

Totally

> Also, this is probably worth bumping the VERSION file but we can do that in another release. I actually wonder if the VERSION should start tracking the current Determined version at the time, so we can roughly correlate them. Even though ideally, we very rarely break the contract of compatibility between the two.

Agree, will do. I'm going to test this some more before landing just incase.

stoksc:
@ioga has a branch that handles this and more.

PR #90, opened 2021-04-29
chore: switch to upstream horovod and NCCL.

ioga:
Switch to upstream horovod and NCCL, but also nail down their versions so they don't auto-upgrade from under us.

horovod commit hash is a placeholder. I intend to update it after https://github.com/horovod/horovod/pull/2879 is landed.


mackrorysd (commented):
Overall looks good, but I'm a little nervous about the idea of not being on a "released" version of upstream Horovod. I suppose that it goes through all of our testing regardless, and I believe they have pretty good CI, and furthermore I don't even think we were sticking to "releases" before, but it gives me pause so I just wanna sleep on it.

VERSION
```diff
@@ -1 +1 @@
-0.12.0
+0.13.0
```
mackrorysd, line 1: Good...

mackrorysd (dismissed):

ioga:
I've noticed NCCL install stuff was a noop because it's already installed in the original image. This allows us to shave off some redundant lines of code.

mackrorysd (approved):
Looks good to me. But let's not merge this until we've tested local builds, so we don't have a repeat of the situation where a user found an old undocumented Docker image that we knew didn't pass our tests.

ioga:
test builds
tf2: `determinedilia/environments:cuda-11.0-pytorch-1.7-lightning-1.2-tf-2.4-gpu-b2592fd`
tf1: `determinedilia/environments:cuda-10.2-pytorch-1.7-tf-1.15-gpu-b2592fd`

PR #94, opened 2021-05-24
feat: CUDA11.2 + Tensorflow 2.5 image.

ioga:
APEX is not happy about minor version mismatch between pytorch and installed CUDA, but they both link to `libcudart.so.11.0`, and there're reports upstream that CUDA 11.2 is very much compatible with 11.0. We apply a small patch to disable this check.


ioga:
Test branch in the main repo: https://app.circleci.com/pipelines/github/determined-ai/determined/12569/workflows/1ee48b37-f4bf-4a80-aff9-24b24b78e072
Nightly tests run in the same time as before; however, currently there's a 1 failure from Anda's recent addition.

.circleci/config.yml
```diff
@@ -102,7 +102,7 @@ workflows:
           context: determined-production
           matrix:
             parameters:
-              image-type: ["tf1-cpu", "tf2-cpu", "tf1-gpu", "cuda-11"]
+              image-type: ["tf1-cpu", "tf2-cpu", "tf1-gpu", "cuda-11", "tf25-gpu"]
```
mackrorysd, line 105: Now that it's the default, how about we start calling the cuda-11 image the -tf2 image? I could do that in a separate change. There's no point in separate CUDA 11 tests anymore either.

ioga: I agree, `tf2-gpu`.

mackrorysd (changes requested):
We'll need to bump the version, too, otherwise it'll overwrite the one we're about to release.

mackrorysd (approved):
Looks good to me. From offline discussion: we're making this a separate image since we're about to make TF2 the default and there are some big changes in here we'd like to keep separate from our defualt.

When we put this in the main repo we should add a peer to the -cuda-11 testing jobs to cover this.

PR #95, opened 2021-06-01
feat: add horovod to CPU images.

ioga:
This is a part of the upcoming CPU-only distributed training feature. It adds about 20M to the image size.
Also, remove redundant `HOROVOD_WITH_TENSORFLOW` and `HOROVOD_WITH_PYTORCH` arguments on the GPU images, as these are default values anyway.
Intended to be landed after https://github.com/determined-ai/environments/pull/94


mackrorysd (approved):
No need for a version change if this goes in immediately after #94.

PR #96, opened 2021-06-02
chore: separate out minimal / universal base layer in Docker images

mackrorysd:
This splits our Dockerfiles into 2 categories:
* Things that are universally important for core Determined functionality, potentially hard to install, and small (the base).
* Things that many customers may not want (e.g. 

This also renames the "CUDA 11" image to represent the fact that it is now the standard Tensorflow 2 image and not a special CUDA 11 preview like it was initially. There will need to be some corresponding changes in the main repo too. It also removes a cloudpickle dependency (Horovod still depends on it, but we don't use it for PyTorch checkpoints anymore) and bumps the version for the next release.

# Testing

GPU tests have already run. I was having some issues with k8s configuration not consistently picking up my modified images, but re-running now and expect to succeed: https://app.circleci.com/pipelines/github/determined-ai/determined/13643/workflows/b0d2188f-dfde-4621-9c38-a5d870d5aebb.


rb-determined-ai (dismissed):
this is a fantastic idea.

cloud/roles/environments/tasks/main.yml
```diff
@@ -37,10 +37,10 @@
     name: "{{ cpu_tf2_environment_name }}{{ image_suffix }}"
     source: pull
 
-- name: Pull CUDA 11 environment
+- name: Pull TF2 GPU environment
   become: yes
   docker_image:
-    name: "{{ cuda_11_environment_name }}{{ image_suffix }}"
+    name: "{{ cpu_tf2_environment_name }}{{ image_suffix }}"
```
ioga, line 43:
```suggestion
    name: "{{ gpu_tf2_environment_name }}{{ image_suffix }}"
```

ioga (dismissed):

cloud/roles/environments/tasks/main.yml
mackrorysd: Good catch, thank you! Always hard to deal with the problems that come up with this post-publish anyway!

mackrorysd:
Using admin to merge, because Ilia +1'd, but then I accepted his suggestion and that reset the review.

PR #103, opened 2021-06-22
patch torch-tb-profiler plugin

azhou-determined:
## Description
including a patched version of torch-tb-profiler plugin in the CPU image which will be used by tensorboard startup

## Checklist

### DIFF
--- a/dockerfile_scripts/torch-tb-profiler-patch.sh
+++ b/dockerfile_scripts/torch-tb-profiler-patch.sh
@@ -4,10 +4,11 @@ set -e
 
 if [ "$TORCH_PROFILER_GIT" ]; then
   TORCH_PROFILER_DIR=/tmp/kineto
-  TORCH_TB_PROFILER_VER="v0.2.0"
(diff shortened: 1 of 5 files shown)


dockerfile_scripts/torch-tb-profiler-patch.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ "$TORCH_PROFILER_GIT" ]; then
+  TORCH_PROFILER_DIR=/tmp/kineto
+  TORCH_TB_PROFILER_VER="v0.2.0"
```
ioga, line 7: optional style preference: you could use the same bash splitting as in apex to allow having this version set in dockerfile instead of here (https://github.com/determined-ai/environments/blob/master/dockerfile_scripts/install_apex.sh#L8)

ioga (approved):

ioga (approved):

PR #105, opened 2021-06-23
fix: another patch for tb-profile-plugin

azhou-determined:
## Checklist

### DIFF
--- a/dockerfile_scripts/torch_tb_profiler_patch.patch
+++ b/dockerfile_scripts/torch_tb_profiler_patch.patch
@@ -1,8 +1,8 @@
 diff --git a/tb_plugin/torch_tb_profiler/plugin.py b/tb_plugin/torch_tb_profiler/plugin.py
-index fa15705..e8a2420 100644
+index fa15705..b882496 100644
 --- a/tb_plugin/torch_tb_profiler/plugin.py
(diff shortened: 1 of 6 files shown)


dockerfile_scripts/torch_tb_profiler_patch.patch
```diff
@@ -13,4 +13,9 @@ index fa15705..e8a2420 100644
 +
                  except Exception as ex:
                      logger.warning("Failed to scan runs. Exception=%s", ex, exc_info=True)
++                    if not self._is_active:
```
ioga, line 16: perhaps a `finally: ` clause will be more appropriate in this try, whatever it's doing?

ioga (approved):

dockerfile_scripts/torch_tb_profiler_patch.patch
azhou-determined: done

ioga: I believe the upper if part can be removed then?

azhou-determined: not quite. there's no exception in that case. it just runs through the for loop, doesn't find anything, and will try again until the end of time.

ioga (approved):

PR #106, opened 2021-06-29
fix: always apply security upgrades to base images.

ioga:
## Description
Nvidia NGC wants our images to have no high severity vulnerabilities. We are currently exposed to https://ubuntu.com/security/CVE-2021-3449
Let's automatically run security upgrades only (via `unattended-upgrades`). These are supposed to be low-risk upgrades.

## Checklist

### DIFF
--- a/Dockerfile-base-cpu
+++ b/Dockerfile-base-cpu
@@ -26,6 +26,8 @@ RUN apt-get update \
 		openssh-server \
 		pkg-config \
 		wget \
+		unattended-upgrades \
+	&& unattended-upgrade \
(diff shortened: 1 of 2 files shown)


mackrorysd (approved):

azhou-determined:
how long do these upgrades usually take and how often are they run? can/should we configure it somehow?

ioga:
these upgrades run just once when images are built and take an extra minute or so. I don't think we should make images upgrade themselves at runtime because 
a) it'll mean extra `apt-get update` at runtime, plus install which may be more than a minute - a significant penalty for each run 
b) task runners may have no internet access

PR #109, opened 2021-07-06
Install recent NCCL version from source and harmonize Horovod versions

mackrorysd:
## Description

When we switched to using upstream NCCL and Horovod we effectively reverted to whatever was installed in the base images, and surprisingly the CUDA 11.1 image is on NCCL 2.7.4, much older than what we were shipping before and what's in the CUDA 10.2 images. The CUDA repositories are not installed by default and those repositories don't include a build of every NCCL version for every CUDA minor version. So we'll go back to building from NVIDIA's upstream source (but pinning to a version). This at least gets us something that is consistent - we may want to make this conditional on what version is already installed in the image at some point.

## Checklist


Dockerfile-default-gpu
```diff
@@ -29,15 +29,20 @@ ARG APEX_GIT
 ARG APEX_PATCH
 RUN /tmp/det_dockerfile_scripts/install_apex.sh
 
+RUN /tmp/det_dockerfile_scripts/build_nccl.sh
+ARG HOROVOD_NCCL_LINK=STATIC
+ARG HOROVOD_NCCL_HOME=/tmp/det_nccl/build
```
ioga, line 34: did we ever confirm that this actually makes horovod use the NCCL we've just built?

mackrorysd: Yes - it's the newer version, AND the CUDA version is an exact match (which is exactly why I wanted to avoid the packages from the repositories).

PR #113, opened 2021-09-10
feat: add tf25-cpu & tf26-cpu images, stop pulling non-default images in the AMIs. [DET-5877]

ioga:
## Description

- There were several requests for CPU images for these TF versions in OSS slack.
- We'll stop pulling in non-default images, i.e. only pull tf 2.4, don't pull tf 1.15, 2.5, or 2.6. (AMIs are getting too big and slow to build)

## Checklist


Makefile
```diff
@@ -73,6 +77,48 @@ build-tf2-cpu:
...
+		--build-arg PYTHON_VERSION="3.7.10" \
+		-t $(DOCKERHUB_REGISTRY)/$(CPU_TF25_BASE_NAME)-$(SHORT_GIT_HASH) \
+		-t $(DOCKERHUB_REGISTRY)/$(CPU_TF25_BASE_NAME)-$(VERSION) \
+		.
+	docker build -f Dockerfile-default-cpu \
+		--build-arg BASE_IMAGE="$(DOCKERHUB_REGISTRY)/$(CPU_TF25_BASE_NAME)-$(SHORT_GIT_HASH)" \
+		--build-arg TENSORFLOW_PIP="tensorflow-cpu==2.5.1" \
+		--build-arg TORCH_PIP="torch==1.7.1 -f https://download.pytorch.org/whl/cpu/torch_stable.html" \
+		--build-arg TORCHVISION_PIP="torchvision==0.8.2 -f https://download.pytorch.org/whl/cpu/torch_stable.html" \
+		--build-arg LIGHTNING_PIP="pytorch_lightning==1.2.0" \
```
shiyuann, line 93: should it be 
```suggestion
		--build-arg TORCH_PIP="torch==1.9 -f https://download.pytorch.org/whl/cpu/torch_stable.html" \
		--build-arg TORCHVISION_PIP="torchvision==0.10.0 -f https://download.pytorch.org/whl/cpu/torch_stable.html" \
		--build-arg LIGHTNING_PIP="pytorch_lightning==1.3.5" \
```
?

Makefile
```diff
@@ -73,6 +77,48 @@ build-tf2-cpu:
...
+		--build-arg PYTHON_VERSION="3.7.10" \
+		-t $(DOCKERHUB_REGISTRY)/$(CPU_TF26_BASE_NAME)-$(SHORT_GIT_HASH) \
+		-t $(DOCKERHUB_REGISTRY)/$(CPU_TF26_BASE_NAME)-$(VERSION) \
+		.
+	docker build -f Dockerfile-default-cpu \
+		--build-arg BASE_IMAGE="$(DOCKERHUB_REGISTRY)/$(CPU_TF26_BASE_NAME)-$(SHORT_GIT_HASH)" \
+		--build-arg TENSORFLOW_PIP="tensorflow-cpu==2.6.0" \
+		--build-arg TORCH_PIP="torch==1.7.1 -f https://download.pytorch.org/whl/cpu/torch_stable.html" \
+		--build-arg TORCHVISION_PIP="torchvision==0.8.2 -f https://download.pytorch.org/whl/cpu/torch_stable.html" \
+		--build-arg LIGHTNING_PIP="pytorch_lightning==1.2.0" \
```
shiyuann, line 114: same.

Makefile
ioga: torch 1.9 doesn't work on cuda 11.2, it has some odd crashes in some of our tests. Me and @azhou-determined didn't have luck in debugging that. So we've got to use CUDA 11.1 for it (which is what we use in tf-2.4 images). torch 1.7 works fine.

But for tf 2.5 and tf 2.6 we need cuda 11.2, so we have to use it, and as a result these "new" tensorflow images end up using "legacy" torch.

shiyuann (approved):

PR #118, opened 2021-09-22
fix: vulnerabilities 2021-09-22.

ioga:
## Description

- Python 3.8.11 for newer images. NGC wants 3.7.12 which is not provided by conda. 3.9.x is not supported by TF2.4.
- Pull in an update in `pillow`.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,15 +6,18 @@ SHORT_GIT_HASH := $(shell git rev-parse --short HEAD)
 NGC_REGISTRY := nvcr.io/isv-ngc-partner/determined
 export DOCKERHUB_REGISTRY := determinedai
 
-CPU_PREFIX := environments:py-3.7-
+CPU_PREFIX := environments:py-3.8-
+CPU_PREFIX_37 := environments:py-3.7-
 CPU_SUFFIX := -cpu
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -6,15 +6,18 @@ SHORT_GIT_HASH := $(shell git rev-parse --short HEAD)
 NGC_REGISTRY := nvcr.io/isv-ngc-partner/determined
 export DOCKERHUB_REGISTRY := determinedai
 
-CPU_PREFIX := environments:py-3.7-
+CPU_PREFIX := environments:py-3.8-
+CPU_PREFIX_37 := environments:py-3.7-
 CPU_SUFFIX := -cpu
 CUDA_102_PREFIX := environments:cuda-10.2-
```
mackrorysd, line 12: Thinking aloud: it would make the name even longer, but it's probably useful to include the Python version in the GPU images one day. Don't know why we never have, we just replace it with CUDA. Let's get this merged, regardless.

mackrorysd (approved):

Makefile
ioga: yep I thought the same while updating this.

PR #119, opened 2021-09-23
feat: run `packer build -force` in CI to allow reruns on failures.

ioga:
## Description

Packer fails from time to time, and rerunning CircleCI job *does not work* because there're artifacts from the previous run. `-force` can override that. As usual, we've got to use it with care and not rerun published builds.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,7 +89,7 @@ jobs:
       - install-ansible
       - run:
           no_output_timeout: 90m
-          command: make publish-cloud-images
+          command: make publish-cloud-images PACKER_FLAGS="-force"
       - store_artifacts:
           path: /tmp/artifacts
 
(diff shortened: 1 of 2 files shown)


mackrorysd:
>> we've got to use it with care and not rerun published builds

Hmm... yeah that makes me a little nervous. Thinking through this, if someone were to republish stuff, either they intentionally change the version back to a previous release to push a bad image, or there's just a temporary blip in unavailability. I think we should check and see if CircleCI has the ability to block people from rerunning things more than a few days old, otherwise this is fine.

ioga:
>block people from rerunning things more than a few days old

I do not see anything like that.

>edit: well what about the case where we have a change after a release and we forget to bump the version to not collide with the old release? Would it normally fail, but now ignore that?

it uses the git commit id as the id, so there won't be issues with our VERSION bumps. 

I suppose the only guidance is "don't rerun successful runs if they're already published", but to be fair, the same is true for any release in the main repo.

mackrorysd (approved):
Okay I feel better about that. At least with Circle CI's UI it seems you really have to go out of your way to run something other than the most recent couple of things. I think this is safe enough.

PR #131, opened 2022-01-27
feat: add deepspeed docker env

liamcli:
## Description
Adding a deepspeed docker image for our deepspeed support.  I originally wanted to build it into our default tf2 gpu image but deepspeed is not compatible with 3.7 cuda arch.  As part of deepspeed installation, custom ops are prebuilt to reduce overhead for JIT during training.  

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ export GPU_TF1_ENVIRONMENT_NAME := $(CUDA_102_PREFIX)pytorch-1.7-tf-1.15$(GPU_SU
 export CPU_TF2_ENVIRONMENT_NAME := $(CPU_PREFIX)pytorch-1.9-lightning-1.3-tf-2.4$(CPU_SUFFIX)
 export GPU_TF2_ENVIRONMENT_NAME := $(CUDA_111_PREFIX)pytorch-1.9-lightning-1.3-tf-2.4$(GPU_SUFFIX)
 export GPU_DEEPSPEED_ENVIRONMENT_NAME := $(CUDA_111_PREFIX)pytorch-1.9-lightning-1.3-tf-2.4-deepspeed-0.5.10$(GPU_SUFFIX)
+export GPU_DEEPERSPEED_ENVIRONMENT_NAME := $(CUDA_111_PREFIX)pytorch-1.9-lightning-1.3-tf-2.4-deeperspeed$(GPU_SUFFIX)
 export CPU_TF25_ENVIRONMENT_NAME := $(CPU_PREFIX)tf-2.5$(CPU_SUFFIX)
 export GPU_TF25_ENVIRONMENT_NAME := $(CUDA_112_PREFIX)tf-2.5$(GPU_SUFFIX)
 export CPU_TF26_ENVIRONMENT_NAME := $(CPU_PREFIX)tf-2.6$(CPU_SUFFIX)
@@ -135,6 +136,7 @@ build-tf2-gpu:
 		-t $(NGC_REGISTRY)/$(GPU_TF2_ENVIRONMENT_NAME)-$(VERSION) \
 		.
 
+# This builds deepspeed environment off of upstream microsoft/DeepSpeed.
 .PHONY: build-deepspeed-gpu
 build-deepspeed-gpu:
 	# Same base image is used for deepspeed as that for the tf2-gpu image.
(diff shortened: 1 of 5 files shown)


dockerfile_scripts/install_deepspeed.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+
+DEBIAN_FRONTEND=noninteractive apt-get install -y pdsh libaio-dev
+if [ "$DEEPSPEED_PIP" ]; then
```
mackrorysd, line 6: If pdsh and libaio-dev are only required for DeepSpeed, shouldn't they be inside the conditional as well? Also, I have a preference for using the -n operator in this conditional. BASH can always do with being more explicit.

mackrorysd (changes requested):

Makefile
```diff
@@ -134,6 +135,36 @@ build-tf2-gpu:
 		-t $(NGC_REGISTRY)/$(GPU_TF2_ENVIRONMENT_NAME)-$(VERSION) \
 		.
 
+.PHONY: build-deepspeed-gpu
+build-deepspeed-gpu:
+	# Same base image is used for deepspeed as that for the tf2-gpu image.
+	docker build -f Dockerfile-base-gpu \
+		--build-arg BASE_IMAGE="nvidia/cuda:11.1-cudnn8-devel-ubuntu18.04" \
```
mackrorysd, line 142: Can we use nvidia/cuda:11.1-cudnn8-devel-ubuntu20.04 instead? We're currently pretty well standardized on that version.

liamcli: i took the same base image as our tf2-gpu image; it also looks like the other images in the Makefile use ubuntu18.04 unless something changed recently.

dockerfile_scripts/install_deepspeed.sh
ioga: Or perhaps this conditional should be in dockerfile.

dockerfile_scripts/install_deepspeed.sh
```diff
@@ -0,0 +1,9 @@
...
+    python -m  pip install ninja psutil packaging attrdict
```
ioga, line 7: `ninja`, `psutil`, and `packaging` seem to be deepspeed dependencies. so maybe explicitly `pip installing` them is not necessary because they'll be pulled in by deepspeed?
not sure about `attrdict`. why do you need to explicitly install it?

dockerfile_scripts/install_deepspeed.sh
```diff
@@ -0,0 +1,9 @@
...
+    DS_BUILD_OPS=1 python -m  pip install $DEEPSPEED_PIP
```
ioga, line 8: while we do still have `python -m pip install` in a couple places, it's not necessary

```suggestion
    DS_BUILD_OPS=1 pip install $DEEPSPEED_PIP
```

dockerfile_scripts/install_deepspeed.sh
liamcli: i removed the deepspeed dependencies since they'll install when installing deepspeed.  i moved `attrdict` to our `additional_requirements` as a quality of life improvement when working with hyperparameters (we can access them as attributes instead of indexing with dict keys).

dockerfile_scripts/install_deepspeed.sh
liamcli: should i standardize all of the pip install calls in this PR?  what's best practices here?

dockerfile_scripts/install_deepspeed.sh
liamcli: i've moved the conditional to top level dockerfile.

Makefile
mackrorysd: Huh - I thought we had standardized everything. Okay, nevermind. We can update them all at once later - and the next Ubuntu LTS will be out soon anyway.

dockerfile_scripts/install_deepspeed.sh
mackrorysd: I wouldn't worry about changing other stuff - though like updating the Ubuntu version, we should probably do that some other time. I suspect this was necessary at some point because of base systems that had conflicting Python versions like 2/3, Ubuntu/Anaconda, etc. but it shouldn't be now.

Makefile
```diff
@@ -165,6 +167,38 @@ build-deepspeed-gpu:
 		-t $(NGC_REGISTRY)/$(GPU_DEEPSPEED_ENVIRONMENT_NAME)-$(VERSION) \
 		.
 
+# This builds deepspeed environment off of a patched version of EleutherAI's fork of DeepSpeed 
+# that we need for gpt-neox support.
+.PHONY: build-deeperspeed-gpu
```
mackrorysd, line 172: I'm not so sure about the name deeperspeed and it's clarity. Would it be wrong to use EleutherAI's name here?

mackrorysd (approved):
Conditionally approved, but please confirm we really want to use the name DeeperSpeed - is that understood by others? If not, it's an easily missed distinction between that and DeepSpeed.

liamcli:
> Conditionally approved, but please confirm we really want to use the name DeeperSpeed - is that understood by others? If not, it's an easily missed distinction between that and DeepSpeed.

EleutherAI chose to call their fork DeeperSpeed.  I think it's also reasonable to name it gpt-neox-deepspeed.

PR #135, opened 2022-02-17
ci: dev builds for environments.

ioga:
## Description

CircleCI machinery for dev builds of docker and cloud images. docker builds are pushed into `determinedai/environments-dev` repo.

With this, one'll be able to
1. push an upstream branch
2. in circleci, approve docker & (optionally) cloud image build
3. wait for 3 hours
4. run bumpenvs (bumpenvs script will need a minor patch to use `-dev` suffixed job ids, because circleci job names between these workflows have to be unique).

## Checklist


ioga:
@azhou-determined I really don't have any good ideas on how to organize that with cirlceci pipelines.

azhou-determined:
> 

@ioga maybe there's a way each build can be its own job that is approved in circleCI? it would require some more code changes to handle configurations like this, but i think the most painful part of dev builds for env images is the waiting 3+ hours. 
still, i think this is helpful for doing a system-wide test of new images. approved.

azhou-determined (approved):

ioga:
@azhou-determined 
"3 hours" take into account the cloud image creation. today cloud image build waits because it needs to cache the default images - if we had an option to do a build without this caching, then it all can be done in parallel, and then the entire thing would be done in 1 hr or 1:30.

mackrorysd (commented):
Does this not result in us pushing Docker images up using whatever the current contents of VERSION is? Hopefully that always refers to the NEXT version and not the current release, but that has bother me that I think our Docker images are pushed up under a version before we actually do the testing and changing code to use it as a default. I don't think this avoids that problem, right?

edit: to be clear, I don't think this patch makes that situation worse, but I just want to make sure we're all on the same page - we need to be very careful not to overwrite the production Docker images with a careless test. We need to always bump VERSION right after a release, and be cognizant of what we're pushing leading up to a release.

ioga:
> Does this not result in us pushing Docker images up using whatever the current contents of VERSION is? Hopefully that always refers to the NEXT version and not the current release, but that has bother me that I think our Docker images are pushed up under a version before we actually do the testing and changing code to use it as a default. I don't think this avoids that problem, right?
> 
> edit: to be clear, I don't think this patch makes that situation worse, but I just want to make sure we're all on the same page - we need to be very careful not to overwrite the production Docker images with a careless test. We need to always bump VERSION right after a release, and be cognizant of what we're pushing leading up to a release.

No, the code sets `REGISTRY_REPO` variable to publish to a different place: https://hub.docker.com/r/determinedai/environments-dev/tags

mackrorysd (approved):

PR #143, opened 2022-03-25
perf: add libibverbs1 to GPU builds

coreystatendet:
## Description

NCCL will not make use of Infiniband RDMA unless `libibverbs1` is installed.  Instead, it will revert to socket-based communication.  Install size is <1MB, so figured we can add to all GPU image builds to support Infiniband RDMA by default.

## Checklist

- [X] Licenses should be included for new code which was copied and/or modified from any external code.


coreystatendet:
At Vishnu's suggestion, also adding `librdmacm1` and `ibverbs-providers` since they are part of the stock horovod Dockerfile here: https://github.com/horovod/horovod/blob/master/docker/horovod/Dockerfile#L44-L46.  Total install size remains < 1MB.

coreystatendet:
See `determined` CI checks here: https://app.circleci.com/pipelines/github/determined-ai/determined/20946/workflows/0684f445-86a6-40fe-a795-c10792b0d445.  I tried to run all that would be relevant for GPU images, let me know if I missed any.

Note that `test-e2e-gke-parallel-tfonly-2` failed above for what seem to be unrelated reasons.  I've been waiting on this PR for a while since a different CI check was failing for unrelated reasons, so instead of delaying further I think it's ok to just look into that separately?

ioga (approved):

PR #145, opened 2022-05-05
Subject: Update PyTorch from 1.10 to 1.11

cdrickett:
## Description

This PR updates the version of PyTorch for the tf2-cpu and tf2-gpu
targets from 1.10 to 1.11. This was done because after building the
tf2-cpu image and converting it to Singularity, a simple PyTorch test
failed with an error similar to:

```
Singularity> python -c "import torch; print(torch.__version__)"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/conda/lib/python3.8/site-packages/torch/__init__.py",
  line 29, in <module>
    from .torch_version import __version__ as __version__
  File
  "/opt/conda/lib/python3.8/site-packages/torch/torch_version.py",
  line 3, in <module>
    from pkg_resources import packaging  # type: ignore[attr-defined]
  File
  "/opt/conda/lib/python3.8/site-packages/pkg_resources/__init__.py",
  line 3267, in <module>
    def _initialize_master_working_set():
<...snip...>
  line 2780, in _get_metadata
    for line in self.get_metadata_lines(name):
  File
  "/opt/conda/lib/python3.8/site-packages/pkg_resources/__init__.py",
  line 1911, in get_metadata_lines
    return yield_lines(self.get_metadata(name))
  File
  "/opt/conda/lib/python3.8/site-packages/pkg_resources/__init__.py",
  line 1898, in get_metadata
    with io.open(self.path, encoding='utf-8', errors="replace") as f:
PermissionError: [Errno 13] Permission denied: '/opt/conda/lib/python3.8/site-packages/.wh.setuptools-58.0.4-py3.8.egg-info'
```
It seemed that the install of a different package in the image
installed a newer setuptools, which caused the conflict. Updating
PyTorch to 1.11 fixed this, as shown below:
```
Singularity> python -c "import torch; print(torch.__version__)"
1.11.0+cpu
```
Tested that the tf2-cpu and tf2-gpu targets build and did simple tests
with PyTorch and TF with each to verify both work.


stoksc:
@cdrickett can we close this one?

PR #146, opened 2022-05-23
chore: security updates for tf 2.6, 2.7, 2.8 [DET-7325]

NicholasBlaskey:
## Checklist

- [X] Licenses should be included for new code which was copied and/or modified from any external code.


ioga:
1. Tensorflow has [released another set of hotfixes](https://github.com/tensorflow/tensorflow/releases) earlier today. We should upgrade straight to these.
2. `VERSION` needs to be bumped to the release which will include this change. I believe `0.18.1` is the one coming out today, so the bump should be to `0.18.2`
3. have you tested this? We've got instructions [in slack here](https://determined-ai.slack.com/archives/CSLAGUF3M/p1647028306535219) and I can provide extra color if necessary.

ioga:
alright the transformers issue should be fixed in determined-ai/determined#4208

PR #148, opened 2022-06-07
multiplatform build with tf2 DET-6296

mpkouznetsov:
## Build multi-architecture images of environments that could theoretically be run on a Mac with M1 or Intel processor.

https://determinedai.atlassian.net/browse/DET-6296

### Problem
Currently, Mac users with Apple Silicon (M1) processor have no official usable environment image. Running experiments on a Mac with M1 requires a `linux/arm64/v8` docker image. (I crafted one for the growing M1 community at HPE mkouznetsov/environments:py-3.8-pytorch-1.9-lightning-1.3-tf-2.7-cpu-4537b95 but it is now 4 months old). The docker images built for x86_64 can be loaded on M1 in emulation mode but crash when certain operations are executed. See, for example, https://github.com/tensorflow/tensorflow/issues/52845

### Solution
Build a multi-platform image.
- A brief explanation of how it could be done: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
- It looks like we have to do it "the hard way": build and publish 2 platform-specific images, then publish a manifest referencing both of them.
- I use https://github.com/KumaTea/tensorflow-aarch64/releases for TF on arm64/v8. TensorFlow is the main reason why non-ARM-specific environment images crash.
- I only build a multi-platform image with TF 2.8. This is currently our default so this is what a Mac user will automatically be using. We are not supporting older versions of TF at this time.

About older TF versions: https://github.com/KumaTea/tensorflow-aarch64/releases states that their TF 2.7 image has an issue: "`Warning: tensorflow_io_gcs_filesystem and tensorflow_io is corrupted and only used for installing tensorflow.`" My hand-rolled environment image mentioned above was built with their TF 2.7 so it would probably work unless one uses GCS.

## Checklist


ioga:
>we have to decide whether we call the image cpu or mp (for multi-platform). A cpu would replace the existing CPU image whereas an mp would exist alongside it.

I'd say it'd be cool to call it `-cpu`, i.e. make the default CPU image be multi-platform. Then M1 users won't need to switch the image locally, and it'd "just work" transparently.

.circleci/config.yml
```diff
@@ -177,6 +257,19 @@ workflows:
       - request-publish-dev-cloud:
           type: approval
           filters: *upstream-feature-branch
+      
+      - build-and-publish-docker-arm:
```
ioga, line 261: have you had a chance to try cross-compilation like what is mentioned here: 
https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
https://blog.siposdani87.com/posts/multi-platform-docker-image-build-on-aarch64.html 
?
that might save us some CI code and build both things together.

In these examples they use go builds, which can totally cross-compile the code. In this case you get the python for the target architecture... and then it has to run to do the work. I suppose that might be a problem, because there's no virtualization layer?..

Dockerfile-default-cpu
```diff
@@ -22,6 +22,7 @@ ARG HOROVOD_WITH_PYTORCH=1
 ARG HOROVOD_WITHOUT_MXNET=1
 ARG HOROVOD_WITHOUT_MPI=1
 ARG HOROVOD_CPU_OPERATIONS=GLOO
+RUN pip install cmake==3.22.4
```
ioga, line 25: does it require this new version? if so, 
1. should we skip `apt-get install`ing cmake in `Dockerfile-base-*`
2. can we install the same cmake version in both `-cpu` and `-gpu` so they match

.circleci/config.yml
```diff
@@ -95,7 +95,62 @@ jobs:
...
+        type: string
+      dev-mode:
+        type: boolean
+        default: false
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 19.03.12
```
ioga, line 110: why is it `setup_remote_docker` here, and `install-docker` in the old build?

ioga (approved):

VERSION
```diff
@@ -1 +1 @@
-0.18.2
+0.18.3
```
ioga, line 1: 0.18.3 is going to be a hotfix patch, so this'd probably go out in 0.18.4

.circleci/config.yml
mpkouznetsov: Yes, I tried it, and ran into problems cross-compiling TF2 for aarch64 on x86_64/amd64. Which is why I am doing ARM image building on ARM hardware.

Dockerfile-default-cpu
mpkouznetsov: 👍

.circleci/config.yml
mpkouznetsov: This job is only publishing to docker repositories.
The other jobs use a `machine` executor and manually install docker on it using `install-docker` - not sure why.
For the ARM image, I think it actually makes sense to use a `machine` executor as it allows me to specify an ARM machine and build natively, without emulation. As I mentioned in my response to another one of your comments, I have been unable to build multi-platform images at once with `buildx` due to `qemu` issues.

mpkouznetsov:
This is superseded by https://github.com/determined-ai/environments/pull/150 : with cross-compilation we can build a multi-platform image in one invocation of `docker buildx build`, obviating the need for a complex job sequencing we did here.

PR #157, opened 2022-07-27
feat: ROCM 5.0 

azhou-determined:
## Description
Support for ROCM 5.0

## Checklist


Makefile
```diff
@@ -42,7 +44,9 @@ export CPU_TF26_ENVIRONMENT_NAME := $(CPU_PREFIX)tf-2.6$(CPU_SUFFIX)
 export GPU_TF26_ENVIRONMENT_NAME := $(CUDA_112_PREFIX)tf-2.6$(GPU_SUFFIX)
 export CPU_TF27_ENVIRONMENT_NAME := $(CPU_PREFIX)tf-2.7$(CPU_SUFFIX)
 export GPU_TF27_ENVIRONMENT_NAME := $(CUDA_112_PREFIX)tf-2.7$(GPU_SUFFIX)
+
 export ROCM_TORCH_TF_ENVIRONMENT_NAME := $(ROCM_42_PREFIX)pytorch-1.9-tf-2.5-rocm
```
ioga, line 48: is the old one still useful? maybe remove it?

ioga (approved):

Makefile
azhou-determined: i think we should just leave it for now, since i'm not sure how many users on actually on the latest ROCM

ioga (dismissed):

Makefile
ioga: I suspect about zero

azhou-determined: good enough for me. removed.

ioga (approved):

ioga (approved):

PR #158, opened 2022-07-28
update deepspeed to 0.6.7

liamcli:
## Checklist


Dockerfile-default-gpu
```diff
@@ -32,16 +32,16 @@ RUN if [ -n "${DET_BUILD_NCCL}" ]; then /tmp/det_dockerfile_scripts/build_nccl.s
 ARG HOROVOD_NCCL_LINK=${DET_BUILD_NCCL:+STATIC}
 ARG HOROVOD_NCCL_HOME=${DET_BUILD_NCCL:+/tmp/det_nccl/build}
 
-ARG HOROVOD_PIP=horovod
```
mpkouznetsov, line 35: not 100% sure but won't you be installing package "0" if no HOROVOD_PIP argument is provided? It seems to be benign but weird.

mpkouznetsov (commented):
Overall this makes sense but please resolve conflicts, and let's look at it actually building the environments (I am wondering if it indeed installs "0").

Dockerfile-default-gpu
liamcli: i think you're right, i changed it to ""

mpkouznetsov (dismissed):

PR #159, opened 2022-08-05
add mpi suffix to artifact log names

liamcli:
## Description
We need to differentiate between artifacts built with or without mpi as part of bumpenvs in the main repo.  This PR adds a MPI flag to the artifact path for each job.  

## Checklist


mpkouznetsov:
Looks good.
I am wondering if the artifact naming is not too cryptic (you have to know that `-1` represents _with_). But since it is for our own consumption with bumpenv it's probably ok.

mpkouznetsov (approved):

PR #171, opened 2022-12-08
chore: Remove tf25 and tf26 builds

drh-determined-ai:
## Description

chore: Remove tf25 and tf26 builds

## Checklist


mpkouznetsov:
I forgot to mention: you also want to update VERSION to contain `0.19.9` instead of `0.19.4` because that is the next version of determined as indicated by https://github.com/determined-ai/determined/blob/master/VERSION

mpkouznetsov (approved):

PR #175, opened 2023-01-12
chore: Remove tensorflow 2.4 recipes

drh-determined-ai:
## Description
https://determinedai.atlassian.net/browse/MLG-245

## Checklist

- [ ] Bump VERSION to make the pushed images are tagged with the right version.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

PR #176, opened 2023-01-13
Add ssh-rsa hostkey workaround

dannysauer:
## Description

Inspired by https://github.com/hashicorp/packer-plugin-ansible/issues/69#issuecomment-1342585096, except our version of packer doesn't seem to yet support the `ansible_ssh_extra_args` parameter (at least, it fails when I use it).  The first two args are for rsa hostkeys, and the "identities only" is what the default value is.

We should probably upgrade packer and use the ssh_extra_args instead.  Someday.

https://developer.hashicorp.com/packer/plugins/provisioners/ansible/ansible#ansible_ssh_extra_args

## Checklist

### DIFF
--- a/cloud/environments-packer.json
+++ b/cloud/environments-packer.json
@@ -32,7 +32,8 @@
         "-e cpu_tf2_environment_name={{ user `docker_registry` }}/{{ user `cpu_tf2_environment_name` }}",
         "-e gpu_tf2_environment_name={{ user `docker_registry` }}/{{ user `gpu_tf2_environment_name` }}",
         "-e short_git_hash={{ user `short_git_hash` }}",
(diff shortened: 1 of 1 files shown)


rb-determined-ai:
> Bump VERSION to make the pushed images are tagged with the right version

What is the risk involved with not bumping the version?  This doesn't seem like it should be changing the content of the images.

dannysauer:
I just checked it because it wasn't applicable.  I'm a fan of checklists being complete.

rb-determined-ai:
got it

mpkouznetsov:
I am guessing this can be closed now that it is part of another PR that has been merged.

dannysauer:
This is in #175. Closing.

PR #177, opened 2023-01-18
chore: update pytorch and tensorflow versions in TF2 images [MLG-14]

drh-determined-ai:
chore: update pytorch and tensorflow versions in TF2 images

## Checklist


mpkouznetsov (approved):
Looks good.

PR #179, opened 2023-01-23
chore: install latest version of torch-tb-profiler (revert 744337d) [MLG-192]

mpkouznetsov:
## Description

Try simply reverting https://github.com/determined-ai/environments/commit/744337d73407826227737fac4b239105c09ea9a3 (from https://github.com/determined-ai/environments/pull/103) to check if the last version of torch profiler plugin works.

## Checklist

## How this should be tested.
1. Modify a PyTorchTrial-base experiment to perform profiling.
### Reproduction
1. Override the environment with one built with the old version of the plugin: determinedai/environments-dev:py-3.8-pytorch-1.12-tf-2.8-cpu-943d0d1
2. Run the experiment. Make sure the profiling is performed (e.g. confirm that the .pt.trace.json files are written)
3. Start a tensorboard also using the environment with the old version of the plugin. Try to open it. Try to navigate to the PyTorch Profiler section. I am not sure what the hang looks like but I guess you will know it when you see it; probably similar to what is described in https://github.com/pytorch/kineto/issues/338 .
### Testing
Same as above but using an environment with the new version: determinedai/environments-dev:py-3.8-pytorch-1.12-tf-2.8-cpu-41040b4

### Examples
If you were to use the [Torch Profiler](https://pytorch.org/docs/stable/profiler.html) in your model definition code as well as a Docker image built with an up-to-date version of `torch-tb-profiler` (v0.4.1 at the time of this writing), you could expect to see something like the following (after the corresponding Determined experiment has completed):
screenshot_tensorboard_profiler_expected

If you were to attempt the same thing but with `torch-tb-profiler` v0.2.0 and _unpatched_, you would see TensorBoard hang indefinitely:
screenshot_tensorboard_old_profiler_unpatched

For comparison, if you were to not use the Torch Profiler, you would find an "empty but complete" page like so:
screenshot_tensorboard_no_profiling


Dockerfile-default-cpu
```diff
@@ -9,9 +9,9 @@ ARG TARGETPLATFORM
 ARG TENSORFLOW_PIP
 ARG TORCH_PIP
 ARG TORCHVISION_PIP
-ARG TORCH_PROFILER_GIT
 RUN if [ "$TENSORFLOW_PIP" ]; then \
-        if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
+        conda install h5py \
+        && if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
             TENSORFLOW_AARCH64_PIP=$(printf '%s' "$TENSORFLOW_PIP" | sed 's/cpu/aarch64/') \
```
mpkouznetsov, line 15: Let's replace conda with `pip install h5py==3.7.0` ?
Also let's move this install inside the AARCH64 clause because I think that is the only architecture that has a problem with h5py.

mpkouznetsov: Actually, now it looks like they belatedly built the binary for aarch64 and uploaded it to [PyPI](https://pypi.org/project/h5py/#files).

image

drh-determined-ai (approved):

mpkouznetsov:
As I said in my previous comment, since they added the binary for aarch64, the whole dance with HDF5 is now unnecessary, and you can revert your last commit.

Makefile
```diff
@@ -193,6 +193,7 @@ DEEPSPEED_VERSION := 0.7.0
 export GPU_DEEPSPEED_ENVIRONMENT_NAME := $(CUDA_113_PREFIX)pytorch-1.10-tf-2.8-deepspeed-$(DEEPSPEED_VERSION)$(GPU_SUFFIX)
 export GPU_GPT_NEOX_DEEPSPEED_ENVIRONMENT_NAME := $(CUDA_113_PREFIX)pytorch-1.10-tf-2.8-gpt-neox-deepspeed$(GPU_SUFFIX)
 export TORCH_PIP_DEEPSPEED_GPU := torch==1.10.2+cu113 torchvision==0.11.3+cu113 torchaudio==0.10.2+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html
+export TORCH_TB_PROFILER_PIP := torch-tb-profiler
```
azhou-determined, line 196: we should probably pin the version of `torch-tb-profiler` since we're pinning its dependencies (torch/torchvision), and this profiler plugin is very prone to breakage.

mpkouznetsov: Done!

drh-determined-ai: I agree with this wholeheartedly.

azhou-determined (approved):
lgtm

PR #180, opened 2023-01-26
fix: torch version in image name

mpkouznetsov:
## Description

TORCH_VERSION was defined after the definition of image names for images containing tensorflow 2.7. As a result, images had an incorrect name, like `docker pull determinedai/environments:cuda-11.2-pytorch--tf-2.7-gpu-mpi-0.19.10`.

Move the definition above its first use in the Makefile.

Previous build: 
image

This branch:
image

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,6 +249,7 @@ define CPU_TF27_TAGS
 endef
 endif
 
+TORCH_VERSION := 1.12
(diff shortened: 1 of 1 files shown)


azhou-determined (approved):
lgtm

PR #181, opened 2023-01-27
Repro tb error with old torch tb profiler

mpkouznetsov:
## Description
This change is not intended to be merged. The intent is to reproduce the tensorboard loading error observed when using torch-tb-profiler plugin version 0.2.0.

## Checklist


mpkouznetsov:
Testing is done. We don't need this anymore.

PR #186, opened 2023-02-22
feat: Add pytorch-only image support

MikhailKardash:
## Description
Add new recipes in Makefile for PT-only image
Add new image types to CircleCI publish-dev workflow
Update Dockerfiles 
Add longer wait time for GCP

## Checklist


Dockerfile-default-gpu
```diff
@@ -9,7 +9,13 @@ ARG TENSORFLOW_PIP
 ARG TORCH_PIP
 ARG TORCHVISION_PIP
 
-RUN if [ "$TENSORFLOW_PIP" ]; then python -m pip install $TENSORFLOW_PIP; fi
+RUN if [ "$TENSORFLOW_PIP" ]; then \
```
mpkouznetsov, line 12: good catch

mpkouznetsov (approved):
LGTM. I would like a second opinion from @ioga

Dockerfile-default-cpu
```diff
@@ -10,12 +10,16 @@ ARG TENSORFLOW_PIP
...
 ARG TORCHVISION_PIP
 RUN if [ "$TENSORFLOW_PIP" ]; then \
+		export HOROVOD_WITH_TENSORFLOW=1 && \
         if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
             TENSORFLOW_AARCH64_PIP=$(printf '%s' "$TENSORFLOW_PIP" | sed 's/cpu/aarch64/') \
             && pip install $TENSORFLOW_AARCH64_PIP; \
         else \
             pip install $TENSORFLOW_PIP; \
         fi; \
+	else \
```
ioga, line 20: I don't think this complex dual variable WITH & WITHOUT logic is necessary.

Looking at the [makefile](https://github.com/horovod/horovod/blob/d15c9147debe3af9929a2c283b5e6d5aefa0cad7/horovod/tensorflow/CMakeLists.txt#L9), if `HOROVOD_WITH_TENSORFLOW=0` and there's no tensorflow present in the image, it'd just build without it.

i.e. we can just have `HOROVOD_WITH_TENSORFLOW` as one ARG.

Makefile
```diff
@@ -342,6 +369,25 @@ build-tf2-gpu: build-gpu-cuda-113-base
...
+		--build-arg APEX_GIT="https://github.com/determined-ai/apex.git@3caf0f40c92e92b40051d3afff8568a24b8be28d" \
+		--build-arg HOROVOD_PIP="horovod==0.24.2" \
+		--build-arg DET_BUILD_NCCL="" \
+		--build-arg HOROVOD_WITH_MPI="$(HOROVOD_WITH_MPI)" \
+		--build-arg HOROVOD_WITHOUT_MPI="$(HOROVOD_WITHOUT_MPI)" \
+		--build-arg HOROVOD_CPU_OPERATIONS="$(HOROVOD_CPU_OPERATIONS)" \
+		-t $(DOCKERHUB_REGISTRY)/$(GPU_PT_ENVIRONMENT_NAME)-$(SHORT_GIT_HASH) \
+		-t $(DOCKERHUB_REGISTRY)/$(GPU_PT_ENVIRONMENT_NAME)-$(VERSION) \
+		-t $(NGC_REGISTRY)/$(GPU_PT_ENVIRONMENT_NAME)-$(SHORT_GIT_HASH) \
+		-t $(NGC_REGISTRY)/$(GPU_PT_ENVIRONMENT_NAME)-$(VERSION) \
```
ioga, line 388: this is not a request for this PR, but I suspect that our publishing to NGC is useless and we could potentially just stop doing it.

Dockerfile-default-cpu
MikhailKardash: Ah, this is true. The [horovod documentation](https://horovod.readthedocs.io/en/stable/install_include.html) was a little confusing. They recommend using WITH = 1 or WITHOUT = 1, but I think using one ARG is definitely cleaner.

Dockerfile-default-cpu
```diff
@@ -31,7 +34,6 @@ ARG TORCH_TB_PROFILER_PIP
 RUN if [ "$TORCH_TB_PROFILER_PIP" ]; then pip install $TORCH_TB_PROFILER_PIP; fi
 
 ARG HOROVOD_PIP=horovod
-ARG HOROVOD_WITH_TENSORFLOW=1
```
ioga, line 34: ideally I'd still like this to be an arg separate from `$TENSORFLOW_PIP`: e.g. if you are building an image off base tensorflow image, you aren't going to have `TENSORFLOW_PIP` set because you will not install it.

however since you'll be installing horovod, you'd want to set this flag.

in other words, it's good we can compute default ARG value based off TENSORFLOW_PIP, but then it'd be great to have a way to override it.

MikhailKardash: Added an override ARG for this.

ioga (approved):

PR #188, opened 2023-03-01
fix: remove pytorch from tf-2.7 image tags [MLG-311]

drh-determined-ai:
## Description

The substring `pytorch-1.12` was in the tags for images built without PyTorch. This mistake was introduced in cd272fd.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -250,9 +250,8 @@ define CPU_TF27_TAGS
 endef
 endif
 
-TORCH_VERSION := 1.12
-export CPU_TF27_ENVIRONMENT_NAME := $(CPU_PREFIX)pytorch-$(TORCH_VERSION)-tf-2.7$(CPU_SUFFIX)
-export GPU_TF27_ENVIRONMENT_NAME := $(CUDA_112_PREFIX)pytorch-$(TORCH_VERSION)-tf-2.7$(GPU_SUFFIX)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -250,9 +250,8 @@ define CPU_TF27_TAGS
 endef
 endif
 
-TORCH_VERSION := 1.12
-export CPU_TF27_ENVIRONMENT_NAME := $(CPU_PREFIX)pytorch-$(TORCH_VERSION)-tf-2.7$(CPU_SUFFIX)
-export GPU_TF27_ENVIRONMENT_NAME := $(CUDA_112_PREFIX)pytorch-$(TORCH_VERSION)-tf-2.7$(GPU_SUFFIX)
+export CPU_TF27_ENVIRONMENT_NAME := $(CPU_PREFIX)-tf-2.7$(CPU_SUFFIX)
```
mpkouznetsov, line 253: This results in `docker pull determinedai/environments-dev:cuda-11.2--tf-2.7-gpu-mpi-0.20.1` (note the double dash). I think (CPU_PREFIX) already contains a trailing dash.

mpkouznetsov: See: https://hub.docker.com/r/determinedai/environments-dev/tags?page=1&name=tf-2.7

drh-determined-ai: Oops, sorry. I should have seen that.

mpkouznetsov (approved):

PR #189, opened 2023-03-03
build: add AWS OFI plugin for NCCL GPU based OFI/libfabric builds

karlonw:
## Description

This PR updates the Dockerfile-base-gpu build target, when using NCCL and OFI, to also build the AWS OFI plugin, which allows NCCL to utilize libfabric as the provider, rather than falling back to TCP when run on Slingshot 11(SS11) systems, which are not IB verbs based. This PR also provides a helper script to the entrypoint, such that if `/host` is bind-mounted into the container, the helper script will attempt to locate the host's libfabric libraries, and prepend them to the `LD_LIBRARY_PATH`  in order to use the most appropriate libfabric for the system.

## Checklist

## Example builds
```
 make  build-tf2-gpu WITH_MPI=1 WITH_OFI=1 WITH_NCCL=1 INTERNAL_AWS_DS=cflhal01.us.cray.com INTERNAL_AWS_PATH=/build_files/nccl
```
```
 make  build-tf2-gpu WITH_MPI=1 WITH_OFI=1 WITH_NCCL=1
```


Nikhil-Nunna (approved):

cdrickett (approved):
Looks good!

Makefile
```diff
@@ -47,9 +62,12 @@ else
 	PLATFORMS := $(PLATFORM_LINUX_AMD_64),$(PLATFORM_LINUX_ARM_64)
 	WITH_MPI := 0
 	OFI_BUILD_ARG := WITH_OFI
+	NCCL_BUILD_ARG := WITH_NCCL
 	HOROVOD_WITH_MPI := 0
 	HOROVOD_WITHOUT_MPI := 1
 	HOROVOD_CPU_OPERATIONS := GLOO
+	HOROVOD_GPU_OPERATIONS := NCCL
+	HOROVOD_GPU_ALLREDUCE :=
```
ioga, line 70: I am a bit worried this line turns off some optimizations for allreduce in the default images.
Also I don't see `HOROVOD_GPU_ALLREDUCE` being set to anything but empty anywhere. why is it necessary?

Dockerfile-default-gpu
```diff
@@ -29,20 +45,27 @@ ARG APEX_GIT
...
+ENV LD_LIBRARY_PATH=${WITH_OFI:+$AWS_LIB_DIR:$LD_LIBRARY_PATH}
 
 ARG HOROVOD_WITH_TENSORFLOW
 RUN if [ "$HOROVOD_WITH_TENSORFLOW" ]; then export HOROVOD_WITH_TENSORFLOW=$HOROVOD_WITH_TENSORFLOW; fi
 
 ARG HOROVOD_PIP=""
 ARG HOROVOD_WITH_PYTORCH=1
 ARG HOROVOD_WITHOUT_MXNET=1
-ARG HOROVOD_GPU_OPERATIONS=NCCL
+#ARG HOROVOD_GPU_OPERATIONS=NCCL
```
ioga, line 64: remove this commented line

jerryharrow (commented):
We also need user documentation as to how to use this image properly.   There is no information anywhere about the -mpi environment images AFAIK.

I can add info in the user docs if I can get a clear description of what this is intended for and the requirements to use it.

* What circumstances would a user choose this image?
* Probably should indicate that these are OpenMP images?
* What needs to be installed on the compute nodes?
* What bind mounts are needed, and when?   Presumably the following for slingshot.  Other cases?

```
bind_mounts:
    - host_path: /usr
      container_path: /host/usr
```

Makefile
karlonw: Initially, I had them both set to NCCL, but got the following build error:
```
      CMake Error at cmake/Utilities.cmake:66 (message):
        Cannot specify both HOROVOD_GPU_OPERATIONS and HOROVOD_GPU_ALLREDUCE
        options.  Try unsetting one of these variables and reinstalling.
      Call Stack (most recent call first): 
```
I will delete the empty `HOROVOD_GPU_ALLREDUCE :=` line

PR #190, opened 2023-03-16
fix: Pin Jupyterlab-server version [MLG-363]

szewaiyuen6:
## Description
This PR is to solve issue ([MLG-363](https://determinedai.atlassian.net/browse/MLG-363)) where Jupyterlab is not launching without internet. This is due to jsonschema library version conflict in the image, thus download of jsonschema package is needed to launch Jupyter lab properly. 

The change pins the package jupyterlab-server, which requires jsonschema. Newer version of jupyterlab-server requires jsonschema to be > 4, which conflicted with requirement of another package :docker-compose . Thus, we are pinning jupyterlab-server version to ensure a single jsonschema version fits all requirements of all packages within the image, thus avoiding the need to download jsonschema. 

## Testing

- Tested that on local Mac (M1), when internet is disconnected, using the image built from this PR to Jupyterlab launches properly
- Tested on AWS cluster (airgapped by ACL inbound and outbound rule), using the image built from this PR to open Jupyterlab works. 
      - Once I applied the ACL rule, I SSH into the master and try to run pip command to download a new library and I saw failure due to network not reachable as expected. This confirms we indeed airgapped the subnet.

### Note
(1) 
For both test cases, we need to open Jupyterlab once first with internet as we need to first download the image we specified. We can then kill the agent instance, and then re-launch Jupyterlab without internet to see if it works as expected.
(2)
The ACL I added does the following:
   - (Inbound )Allow only traffic from my IP to subnet
   - (Outbound) Allow only traffic from subnet to my IP (and not to other IPs)
   - I followed this guide to add ACL: https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html#CreateACL
       - note that ACL needs to be created on a VPC level, then associate with a subnet

## Checklist

[MLG-363]: https://hpe-aiatscale.atlassian.net/browse/MLG-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mpkouznetsov:
1. Do you know whether _just rebuilding_ an environment (without pinning the jupyter-server) would have solved the problem?
2. Please also update VERSION setting it to `0.20.2`. That is our next release version that will be pointing at this.
3. After we land this, please ask me about `bumpenvs` procedure that we have to do as a follow-up to this.

PR #192, opened 2023-03-17
bump version

mpkouznetsov:
## Description

Bump the version.
Also check if the air-gapped Jupyter Lab problem goes away.

## Checklist


mpkouznetsov:
@szewaiyuen6 confirmed that just rebuilding the image does not solve the problem with running air-gapped Jupyter Lab. Closing.

PR #194, opened 2023-04-04
fix: remove tf from DeepSpeed Docker image names [MLG-393]

drh-determined-ai:
## Description

Ever since we started building environment images with for DeepSpeed experiments (Feb 2022, #131), we'd planned to not install Tensorflow. At some point (#158), this installation step was removed from the Docker builds, but the name of the images was not changed. This PR fixes that.

## Checklist


mpkouznetsov:
Could you actually update the VERSION to 0.21.2 .
Then we need to see if the `publish-dev` failure is intermittent, and if it happens again, investigate.

mpkouznetsov (approved):
LGTM

PR #195, opened 2023-04-06
chore: replace pynvml with nvidia-ml-py

ioga:
## Description

[pynvml](https://pypi.org/project/pynvml/) is an older package which is supposed to be identical to [nvidia-ml-py](https://pypi.org/project/nvidia-ml-py/) but it's not really the same: [gpustat](https://github.com/wookayin/gpustat/blob/master/gpustat/nvml.py) does some low level stuff which requires the latter.

this was discovered in determined-ai/determined#6480.

## Checklist


MikhailKardash (approved):
LGTM

PR #196, opened 2023-04-11
TF2_VERSION 2.12

drh-determined-ai:
## Checklist


mpkouznetsov (approved):
This looks good.

mpkouznetsov (approved):

mpkouznetsov:
I don't like the 30 minutes time out on the publish step that much, given that it typically takes a couple of minutes. But we can look at it later.

PR #200, opened 2023-04-18
chore: CircleCi GPU instances, DeepSpeed 0.8.3 [MLG-399]

drh-determined-ai:
## Description

This PR upgrades our Docker images to use DeepSpeed 0.8.3 (up from 0.7.0). A major side-effect of this upgrade is that these images must be built on machine with a GPU present. This required waiting for our contract with CircleCI to renew as yearly, granting us access to their GPU instances. Using these required changes to our CircleCI configuration: we have split the build jobs into two sets, one with the same CPU setup as before and the other - for DeepSpeed and DeepSpeed/NeoX - with a GPU setup. 

These images have already been tested in the draft PR [#6614](https://github.com/determined-ai/determined/pull/6614). Specifically, the job `test-e2e-deepspeed` [succeeded](https://app.circleci.com/pipelines/github/determined-ai/determined/37272/workflows/0e0355d9-5528-44ec-aef8-441b21d2c648/jobs/1352039).

## Checklist

- [ ] Bump VERSION to make the pushed images are tagged with the right version.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

PR #204, opened 2023-04-26
chore: DeepSpeed 0.8.3 [MLG-399]

drh-determined-ai:
## Checklist


tayritenour (approved):
LGTM

drh-determined-ai:
In order for DeepSpeed 0.8.3 to build while building the corresponding Docker images, we not only had to run `docker` on a CircleCI instance with a GPU (this was done in #200) but also install the `nvidia-container-runtime` and hack the Docker daemon to use the "nvidia" runtime: this cannot be done in a script due to lack of privileges (even with `sudo`), which is why it's done in a CircleCI command.

tayritenour (approved):
Is there any additional work we need to do to make sure that the deepspeed tests run with these new images? Otherwise looks good to me

PR #209, opened 2023-05-10
fix: use `CMD /bin/bash` by default.

ioga:
## Description

Since we now set a custom entrypoint script, let's also set command to `/bin/bash` as well to make `docker run -it IMAGE_TAG` work.

## Checklist


karlonw (approved):
Looks good. Sorry I missed this use case earlier.

MikhailKardash:
Probably should make this change in `dockerfile-base-cpu` as well

MikhailKardash (approved):
LGTM

PR #212, opened 2023-05-19
fix: move "determined" to front in nsswitch.conf

dougdet:
## Description

[HELP-51](https://hpe-aiatscale.atlassian.net/browse/HELP-51)

In `nsswitch.conf` move "determined" to the front of the list so Determined-created identity entries are picked up first.  This prevents conflicting entries that may be added by the infrastructure.

In the ticket linked above, distributed training via SSH, e.g. Horovod, on OpenShift 4.+ gets thrown off by [CRI-O injecting a `passwd` entry](https://github.com/cri-o/cri-o/pull/2022/commits/77408ef1490002e62c88baacc5c994e97aa793c6) that interferes with the SSH client identity as expected by Determined.

@dougdet worked with @rb-determined-ai on this solution.

## Checklist

[HELP-51]: https://hpe-aiatscale.atlassian.net/browse/HELP-51?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockerfile_scripts/install_libnss_determined.sh
+++ b/dockerfile_scripts/install_libnss_determined.sh
@@ -5,5 +5,19 @@ set -e
 # Add a plugin to the user system that lets us extend the users available in
 # the container at runtime. This is critical for supporting non-root shell,
 # which in turn is critical for non-root distributed training.
+#
+# Also, to work around a behavior in OpenShift 4.+ [1], place `determined`
+# NSS plug-in to be searched first instead of last so Determined-defined
+# entries are used when there are conflicts with infrastructure-based
+# entries in these identity databases: passwd, shadow, group.
+#
+# [1] https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids "By default,
+# OpenShift 4.x appends the effective UID into /etc/passwd of the Container
+# during the creation of the Pod."
+
 make -C /tmp/det_dockerfile_scripts/libnss_determined libnss_determined.so.2 install \
-    && sed -E -i -e 's/^((passwd|shadow|group):.*)/\1 determined/' /etc/nsswitch.conf
+    && sed -i -E '
+        /^(passwd|group|shadow):/{
+            s/[ \t]determined//;
+            s/:/: determined/;
+        }' /etc/nsswitch.conf
--- a/cloud/environments-packer.json
+++ b/cloud/environments-packer.json
@@ -84,7 +84,7 @@
         "us-east-2"
       ],
       "source_ami": "{{ user `aws_base_image`}}",
-      "instance_type": "p2.xlarge",
+      "instance_type": "g4dn.xlarge",
       "ssh_username": "ubuntu",
       "launch_block_device_mappings": [
         {


dannysauer:
> Funny, dougdet's primary email address is [doug.parrish@hpe.com](mailto:doug.parrish@hpe.com). Figure this would be accepted as a CLA, or something akin to it.

The CLAbot doesn't look up the emails associated with the github username (mostly because that would require extra permissions which github doesn't allow); it only goes off of the email address associated with the commit - which is what's configured in your .gitconfig.  What you likely need is something like
```
git config --global user.email doug.parrish@hpe.com
```
Then update your old commit to use the new author with `git commit --amend` and force-push the branch (`git push --force-with-lease`).  It'll be happier then.

dockerfile_scripts/install_libnss_determined.sh
```diff
@@ -5,5 +5,8 @@ set -e
 # Add a plugin to the user system that lets us extend the users available in
 # the container at runtime. This is critical for supporting non-root shell,
 # which in turn is critical for non-root distributed training.
+# Also, place `determined` NSS plug-in to be searched first instead of last so
+# Determined-defined entries are used when there are conflicts with infrastructure-
+# based entries in these identity databases: passwd, shadow, group
```
rb-determined-ai, line 10: In your comment, can you specify that this specifically fixes OpenShift, and briefly explain why?

dockerfile_scripts/install_libnss_determined.sh
```diff
@@ -5,5 +5,8 @@ set -e
...
 make -C /tmp/det_dockerfile_scripts/libnss_determined libnss_determined.so.2 install \
-    && sed -E -i -e 's/^((passwd|shadow|group):.*)/\1 determined/' /etc/nsswitch.conf
+    && sed -i -E '/^(passwd|group|shadow):/s/[ \t]determined//;/^(passwd|group|shadow):/s/:([ \t]+)/:\1determined /' /etc/nsswitch.conf
```
rb-determined-ai, line 12: What's wrong with the simpler script I suggested?

```bash
sed -i -E 's/^(group|passwd|shadow): /\1: determined /'
```

Yours is complicated enough I'd have a lot of trouble understanding what I was reading, but I'm not sure I why it is better or safer?

In fact... now that I look closer at it, I think the `s/[ \t]determined//;` bit is unnecessary, since that is to delete a `determined` entry, but there shouldn't be one in a vanilla image.  That was something we did when we were modifying Determined images.

dannysauer:
Reposting from Slack:

This feels like a potential security issue. What happens to availability if somebody creates a user in determined named root, for example? How does nsswitch behave if there's an error in the module or if the module is missing?

Local accounts *should* override network accounts by default. If the local infrastructure is creating conflicting accounts which aren't consistent across machines, I'd suggest that's a bigger infrastructure mismanagement issue which probably should be addressed, with the end-user optionally making this change as more of a last-resort than the default.

dockerfile_scripts/install_libnss_determined.sh
dannysauer: Given that it's in single quotes, the one-liner could be split up on the semicolons to improve readability.  And really, the two substitutions should probably be grouped into a block to avoid duplication since they're using the same line selector.

```suggestion
    && sed -i -E '
    /^(passwd|group|shadow):/{
      s/[ \t]determined//;
      s/:([ \t]+)/:\1determined /;
    }' /etc/nsswitch.conf
```

dannysauer: (actually, I'm no longer positive about the `\s` - that may or may not be universal.  I'd probably actually use `[[:space:]]`, but :shrug:

dockerfile_scripts/install_libnss_determined.sh
dougdet: How about the following for a comment section?

```
# Add a plugin to the user system that lets us extend the users available in
# the container at runtime. This is critical for supporting non-root shell,
# which in turn is critical for non-root distributed training.
#
# Also, to work around a behavior in OpenShift 4.+ [1], place `determined`
# NSS plug-in to be searched first instead of last so Determined-defined
# entries are used when there are conflicts with infrastructure-based
# entries in these identity databases: passwd, shadow, group.
#
# [1] https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids "By default,
# OpenShift 4.x appends the effective UID into /etc/passwd of the Container
# during the creation of the Pod."
```

dannysauer:
The [docs](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids) indicate this:

> When the application also needs to be executed under a specific custom UID, then the Pod definition needs to use a ServiceAccount with the RunAsAny privilege and the Pod needs to use the runAsUser to set the UID to the hardcoded UID. NOTE: Using a hardcoded UID is NOT recommended. Among issues with this approach is that it is prone to UID collisions with system UIDs or with UIDs of the same or different application running in a different Namespace expecting to use the same UID.
> For the Pod to be able to specify the UID to use during execution, in OpenShift 4.4, the ServiceAccount must have at least the “nonroot” or “anyuid” SCC’s

Would it perhaps make more sense to just update our pod template when running on OpenShift to work within their system?  Does that negate the need for the nsswitch library on Openshift entirely, or do we still need to support other users in a given pod?

dockerfile_scripts/install_libnss_determined.sh
dougdet: Amending after @dannysauer's use of `{cmd;cmd}` syntax which, until today, I was unware 🤦🏻‍♂️ 

`/^(passwd|group|shadow):/{s/[ \t]determined//;s/:([ \t]+)/:\1determined /}`

1. `/^(passwd|group|shadow):/` Run substitute only the lines beginning with passwd, group, shadow; none of the others
2. `s/[ \t]determined//;` Remove "determined", and leading space|tab, if it happens to already exist
3. `s/:([ \t]+)/:\1determined /` Insert "determined" after all intervening spaces and tabs between the colon after the database name and the first _source_

@rb-determined-ai, the reason for this particular sed script is I get nervous if start accepting too many assumptions about what I'm updating.  If I think of something that can break my algorithm, I try to prevent it provided the solution is reasonable and feasible ... and maintainable.  Danny proposed a readability enhancement without reducing its functionality.

dougdet:
@dannysauer

dockerfile_scripts/install_libnss_determined.sh
```diff
@@ -5,5 +5,19 @@ set -e
...
+# OpenShift 4.x appends the effective UID into /etc/passwd of the Container
+# during the creation of the Pod."
+
 make -C /tmp/det_dockerfile_scripts/libnss_determined libnss_determined.so.2 install \
-    && sed -E -i -e 's/^((passwd|shadow|group):.*)/\1 determined/' /etc/nsswitch.conf
+    && sed -i -E '
+        /^(passwd|group|shadow):/{
+            s/[ \t]determined//;/^(passwd|group|shadow):/;
+            s/:([ \t]+)/:\1determined /;
+        }' /etc/nsswitch.conf
```
rb-determined-ai, line 23: this doesn't look quite right... there seem to be two copies of `/^(passwd|group|shadow):/;` and I don't think they're placed in a particularly meaningful way.

dannysauer: Yeah, there's a selector with no action there - which I'm pretty sure is not the intention. :)

dannysauer:
> Part of the problem was OpenShift appended an entry that had `/`(root) for the home directory, and `/usr/sbin/nologin` for the login shell.

Hmm, that's a solid point.  I dug around some and can't find where they define the template for a new entry.  I guess in typical use for a container those shouldn't matter, but clearly it does matter to us.  I wonder if HPE has a Redhat support contact.  It'd be neat if we could make the default user actually be usable. :)

dockerfile_scripts/install_libnss_determined.sh
dougdet: (groan) Editing error.  Updated to

```
        /^(passwd|group|shadow):/{
            s/[ \t]determined//;
            s/:([ \t]+)/:\1determined /;
        }' /etc/nsswitch.conf
```

dannysauer: I'm clearly partial to my suggestion with the colon inside the capture group instead of repeating the colon in the replacement.  But either way, this looks better.

dougdet: Done.

```
        /^(passwd|group|shadow):/{
            s/[ \t]determined//;
            s/(:[ \t]+)/\1determined /;
        }' /etc/nsswitch.conf
```

dockerfile_scripts/install_libnss_determined.sh
```diff
@@ -5,5 +5,19 @@ set -e
...
+# [1] https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids "By default,
+# OpenShift 4.x appends the effective UID into /etc/passwd of the Container
+# during the creation of the Pod."
+
 make -C /tmp/det_dockerfile_scripts/libnss_determined libnss_determined.so.2 install \
-    && sed -E -i -e 's/^((passwd|shadow|group):.*)/\1 determined/' /etc/nsswitch.conf
+    && sed -i -E '
+        /^(passwd|group|shadow):/{
+            s/[ \t]determined//;
+            s/(:[ \t]+)/\1determined /;
```
rb-determined-ai, line 22: suggested simplification: `s/:/: determined/`

rb-determined-ai (approved):

dockerfile_scripts/install_libnss_determined.sh
dougdet:
```
passwd:         files systemd
group:          files systemd
shadow:         files
gshadow:        files
```

.. becomes

```
passwd: determined         files systemd
group: determined          files systemd
shadow: determined         files
gshadow:        files
```

I really don't mind.  I was merely preserving appearance, but who's gonna be looking at `nsswitch.conf` in an ephemeral container anyway?

dougdet:
Made a change in `cloud/environments-packer.json` to use "gp3" volume type in GovCloud.  Repeat failures in CircleCI when publishing the images due to `amazon-ebs-gov,error,Error launching source instance: InsufficientInstanceCapacity`.

PR #214, opened 2023-05-31
fix: insufficient instance capacity error

mpkouznetsov:
## Description

Use g4dn.xlarge for gov zones.

This is addressing the following error:
```
Error launching source instance: InsufficientInstanceCapacity: We currently do not have sufficient p2.xlarge capacity in zones with support for 'gp2' volumes. Our system will be working on provisioning additional capacity.
```

## Checklist

### DIFF
--- a/cloud/environments-packer.json
+++ b/cloud/environments-packer.json
@@ -114,7 +114,7 @@
         "us-gov-west-1"
       ],
       "source_ami": "{{ user `gov_aws_base_image`}}",
-      "instance_type": "m5.large",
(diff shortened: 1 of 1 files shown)


dougdet:
In PR #212 I put in the same change to `cloud/environments-packer.json`.  Seems more appropriate to have it in its own PR, i.e. this one.

dougdet (approved):
Noticing this instance_type change is being applied to only GovCloud

ioga (approved):

PR #218, opened 2023-07-11
chore: upgrade ubuntu to 22.04 LTS [MLG-695]

mpkouznetsov:
## Description
MLG-695. Customers are asking for Ubuntu 22.04 in images. Ubuntu 22 is 15 months old at this point.

## Checklist


mpkouznetsov:
Not that simple. We would have to upgrade CUDA to at lease 11.7: https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=cudnn8-devel-ubuntu22

PR #221, opened 2023-07-27
Remove tf1 and bugfixes

MikhailKardash:
## Description
remove `tf1` images
bump `tensorflow-cpu==2.8.3` to `tensorflow-cpu==2.8.4` to address a build issue
update conda version to address a build issue

## Commentary
`update-bumpenvs-yaml.py` in master will need to be updated to reflect this change.

## Checklist

- [x] Test the images by running the test `bumpenvs` procedure in the determined repo. See [README](https://github.com/determined-ai/environments/blob/main/README.md).

## Ticket
MLG-666


dockerfile_scripts/install_python.sh
```diff
@@ -8,13 +8,13 @@ ARCHITECTURE=${2:-linux/amd64}
...
-  CONDA_INSTALLER="Miniconda3-py39_4.10.3-Linux-x86_64.sh"
-  CONDA_MD5="8c69f65a4ae27fb41df0fe552b4a8a3b"
+  CONDA_INSTALLER="Miniconda3-py39_23.5.2-0-Linux-x86_64.sh"
+  CONDA_SHA256="9829d95f639bd0053b2ed06d1204e60644617bf37dd5cc57523732e0e8d64516"
   CONDA_URL="https://repo.anaconda.com/miniconda"
 elif [[ "$ARCHITECTURE" == "linux/arm64" ]]; then
-  CONDA_INSTALLER="Miniforge3-4.10.3-10-Linux-aarch64.sh"
-  CONDA_MD5="85d7ea630bb91259bf09f6d52a5ec1c4"
-  CONDA_URL="https://github.com/conda-forge/miniforge/releases/download/4.10.3-10"
+  CONDA_INSTALLER="Mambaforge-23.1.0-4-Linux-aarch64.sh"
```
ioga, line 15: it seems like this mambaforge thing exists for linux/amd64. perhaps we could use different architectures of the same mambaforge installer here, rather than different versions of miniconda and mambaforge?

ioga (approved):

dockerfile_scripts/install_python.sh
MikhailKardash: I wanted to version bump while keeping the previous URL. However the Mambaforge is a bit behind in version from Miniconda. It looks like Miniconda has an in-house `aarch64` installer now. I'll try that instead.

PR #226, opened 2023-09-26
Pad 8

will-HPE:
## Description
feat: add multi-node support to the ROCM image.
build: added new targets to Makefile
build: significant changes to Dockefile-default-rocm
build: added new support scripts (build_aws_rocm.sh and ompi_rocm.sh)
build: large changes to scrape_libs.sh

testing on slurm:
an example command:
srun --mpi=pmi2 -n8 --ntasks-per-node=4 -p bard singularity run --pwd /var/tmp --no-mount=tmp --writable-tmpfs --bind /scratch/testuser/libfabric/:/det_libfabric --bind /:/det_host --rocm w/rocm-5.6-pytorch-2.0-tf-2.10-rocm-0.26.1.sif  python /home/users/testuser/horovod/examples/pytorch/pytorch_synthetic_benchmark.py --batch-size=96 --fp16-allreduce --model=resnet50

further testing:
https://github.com/determined-ai/determined-ee/pull/1067

## Checklist


Dockerfile-default-rocm
```diff
@@ -1,40 +1,138 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
+#RUN sudo sysctl -w kernel.core_pattern="core-%h-%e-%s-%t-%u-%p"
```
karlonw, line 3: guessing some of these commented-out lines were for debugging?

Dockerfile-default-rocm
```diff
@@ -1,40 +1,138 @@
...
+#RUN dpkg -S /opt/ompi
+RUN apt remove -y openmpi ucx
+#RUN mv /opt/ompi /opt/ompi.0 && mv /opt/ucx /opt/ucx.0
+RUN mv /opt/ompi /opt/ompi.0 
+#RUN ln -s /container/ompi /opt && ln -s /container/ucx /opt
+RUN ln -s /container/ompi /opt 
+#RUN ls /opt/ompi && exit 99
 
 RUN mkdir -p /var/run/sshd
+#RUN pip install torch torchvision torchaudio
```
karlonw, line 14: supposed to remove torch?

Dockerfile-default-rocm
```diff
@@ -1,40 +1,138 @@
...
+                libibverbs1 \
+                libkrb5-dev \
+                librdmacm1 \
+                libssl-dev \
+                libtool \
 		git \
 		krb5-user \
 		cmake \
-		g++-4.8 \
+		g++ \
```
karlonw, line 35: should this be locked to a specific version of g++?

Dockerfile-default-rocm
```diff
@@ -1,40 +1,138 @@
...
+                libnuma-dev \
+                libpmi2-0-dev \
 		unattended-upgrades \
 	&& unattended-upgrade \
 	&& rm -rf /var/lib/apt/lists/* \
 	&& rm /etc/ssh/ssh_host_ecdsa_key \
 	&& rm /etc/ssh/ssh_host_ed25519_key \
 	&& rm /etc/ssh/ssh_host_rsa_key
-
+#RUN pip install pip install --upgrade pip
```
karlonw, line 51: more debug statements?

dockerfile_scripts/build_aws_rocm.sh
```diff
@@ -0,0 +1,74 @@
...
+
+  cuda_ver_str=`echo $CUDA_VERSION | awk -F "." '{print $1"."$2}'`
+  CUDA_DIR="/usr/local/cuda-$cuda_ver_str/targets/x86_64-linux"
+  GDRCOPY_HOME="/usr"
+  OFI_INSTALL_DIR=/container/ofi
+  ls $OFI_INSTALL_DIR
+  [ "$WITH_MPICH" = "1" ]&&MPI_INSTALL_DIR=${MPICH_INSTALL_DIR}||MPI_INSTALL_DIR=${OMPI_INSTALL_DIR}
+  AWS_CONFIG_OPTIONS="--prefix ${AWS_PLUGIN_INSTALL_DIR} \
+	  --with-libfabric=${OFI_INSTALL_DIR}            \
+	  --with-rccl=${HOROVOD_NCCL_HOME}               \
```
karlonw, line 41: I can't find where HOROVOD_NCCL_HOME is set, similar to https://github.com/determined-ai/environments/blob/main/Dockerfile-default-gpu#L12

dockerfile_scripts/scrape_libs.sh
```diff
@@ -27,13 +35,11 @@ then
            # to set these paths based on where we find the libs that
            # libfabric is dependent upon.
            export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$libfabric_dir
```
karlonw, line 37: comment above says "prepend" but the line "post-pends" to LD_LIBRARY_PATH

dockerfile_scripts/scrape_libs.sh
```diff
@@ -27,13 +35,11 @@ then
...
-           export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/host/usr/lib64
-           export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/host/usr/local/lib64
        fi # end if found libfabric.so
-   fi # end if /host exists
+   fi # end if /det_libfabric exists
    export LD_LIBRARY_PATH=/container/aws/lib:$LD_LIBRARY_PATH
 
-   tmp_nvcache_dir=$(mktemp -d -p /var/tmp $(whoami)-nvcache-XXXXXXXX)
+   tmp_nvcache_dir=$(mktemp -d -p /var/tmp ${WHOAMI}-nvcache-XXXXXXXX)
```
karlonw, line 42: I don't see where tmp_nvcache_dir is used?

dockerfile_scripts/scrape_libs.sh
```diff
@@ -55,7 +61,51 @@ then
...
       export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libffi.so.7:$LD_PRELOAD
    fi
+# Execute what we were told to execute
+
+   if [ "$WITH_RCCL" = "1" ]; then
+      #export LD_LIBRARY_PATH=/scratch2/${WHOAMI}/lib/libfabric-1.21.1:/scratch2/${WHOAMI}/lib:/container/aws/lib:$LD_LIBRARY_PATH
+      #export LD_LIBRARY_PATH=/scratch2/${WHOAMI}/lib/libfabric-1.21.1:/scratch2/${WHOAMI}/lib:$LD_LIBRARY_PATH
+
+      export TF_FORCE_GPU_ALLOW_GROWTH=true
+      export CUDA_CACHE_PATH=/tmp/${WHOAMI}.nvcache
```
karlonw, line 71: should this be in /vat/tmp/....?

dockerfile_scripts/scrape_libs.sh
```diff
@@ -55,7 +61,51 @@ then
...
+      export CXI_FORK_SAFE=1
+      export FI_EFA_FORK_SAFE=${CXI_FORK_SAFE}
+      # Avoid hang with NCCL/RCCL and libfabric
+      export FI_MR_CACHE_MONITOR=userfaultfd
+
+      export NCCL_DEBUG=TRACE
+      export RCCL_DEBUG=${NCCL_DEBUG}
+
+      if [ "$WITH_NFS_WORKAROUND" = "1" ]; then
+         export MIOPEN_USER_DB_PATH="/tmp/${WHOAMI}_${SLURM_LOCALID}"
```
karlonw, line 100: should this be in /vat/tmp/....?

dockerfile_scripts/scrape_libs.sh
```diff
@@ -1,12 +1,20 @@
 #!/bin/bash
 
+WHOAMI=$(whoami)
 # Only scrape host libs if AWS/NCCL/OFI was built for this image
 if [ -d /container/aws/lib ]
 then
    # See if we mounted in host libs in the expected location
-   host_dir="/host"
+   host_dir="/det_host"
```
karlonw, line 8: not sure I understand the difference in /det_host and /det_libfabric

will-HPE:
Karlon,

What version of this file are referencing?

The code I see is this:
           # Prepend the tmp dir for the host libfabric
           export LD_LIBRARY_PATH=$tmp_lib_dir:$LD_LIBRARY_PATH

will-HPE:
Karlon,

That was put in place because it was retrieving incompatible versions. Hopefully, this can be relaxed at some point.

will-HPE:
This was hopeless attempt capture extra info on core files. Removing.

will-HPE:
Karlon,

tmp_nvcache_dir isn't new to this commit.

will-HPE:
Karlon,

I'm not seeing the code you are referencing:
grep vat dockerfile_scripts/scrape_libs.sh

will-HPE:
Karlon,

It was requested by Jerry for two reasons: firstly, so the libfabric.so find command wouldn't search the whole filesystem; and, when that search returns two libfabric.so's it breaks, said breaking should be prevented by passing a specific directory.

karlonw:
Something’s weird. I sent a screenshot in slack

Dockerfile-default-rocm
```diff
@@ -48,7 +144,10 @@ RUN /tmp/det_dockerfile_scripts/install_google_cloud_sdk.sh
 # google-api-python-client -> google-api-core -> googleapis-common-protos -> protobuf
 # Horovod cannot build with protobuf > 3.20.x
 # latest google-api-python-client requires protobuf >= 3.20.1
-RUN pip install protobuf==3.20.1
+#RUN pip install protobuf==3.20.1
+RUN pip install numpy==1.23.1
+RUN pip install protobuf==3.19.6
```
MikhailKardash, line 149: `pip install numpy` again? and a protobuf install, then a downgrade?

Dockerfile-default-rocm
MikhailKardash: Agreed, all commented out code should be removed.

will-HPE:
There doesn't seem to be a perfect set of package choices that satisfies all dependencies.

This should revisited to see if the landscape has improved.

Dockerfile-default-rocm
```diff
@@ -63,13 +162,87 @@ ARG HOROVOD_WITH_TENSORFLOW=1
...
+#TEST ENVs
+ENV HOROVOD_GPU_ALLREDUCE=$HOROVOD_GPU_ALLREDUCE
+ENV HOROVOD_GPU_BROADCAST=$HOROVOD_GPU_BROADCAST
+
 RUN if [ "$HOROVOD_PIP" != "0" ]; then pip install "${HOROVOD_PIP}" ; fi
 
-RUN rm -r /tmp/*
+RUN pip install -r /tmp/det_dockerfile_scripts/additional-requirements.txt
+
+RUN pip uninstall -y tb-nightly tensorboardX tensorboard
```
MikhailKardash, line 196: `pip uninstall tensorboard` seems unnecessary if we just pip install it later.

will-HPE:
This resolved a specific error; should be revisited to see if it's still an issue.

Dockerfile-default-rocm
```diff
@@ -1,40 +1,124 @@
...
+                ca-certificates \
+                curl \
+                daemontools \
+                glibc-source \
+                ibverbs-providers \
+                libibverbs1 \
+                libkrb5-dev \
+                librdmacm1 \
+                libssl-dev \
+                libtool \
```
ioga, line 27: what's up with this change? it'd be good to figure out tabs vs spaces.

Dockerfile-default-rocm
```diff
@@ -1,40 +1,124 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 
+RUN apt remove -y openmpi ucx
+#Let's move any existing /opt/ompi out of the way
+RUN mv /opt/ompi /opt/ompi.0
```
ioga, line 6: why not just remove it? it is possible to still make the existing `/opt/ompi` work somehow?

dockerfile_scripts/build_aws_rocm.sh
```diff
@@ -0,0 +1,74 @@
...
+  mkdir -p ${AWS_SRC_DIR}                         && \
+    cd ${AWS_SRC_DIR}                             && \
+    git clone https://github.com/ROCmSoftwarePlatform/aws-ofi-rccl && \
+    cd aws-ofi-rccl                               && \
+    ./autogen.sh                                  && \
+    CC=hipcc ./configure ${AWS_CONFIG_OPTIONS}    && \
+    make                                          && \
+    make install                                  && \
+    cd /tmp                                       && \
+    rm -rf ${AWS_SRC_DIR}
```
ioga, line 73: why are these scripts written like this? if the purpose is to err out on any error, can we just `set -e`?

ioga (commented):
general question: is there an existing image out there which we can reuse and install determined addons into, rather then rebuilding the whole thing ourselves?

Dockerfile-default-rocm
```diff
@@ -63,13 +149,71 @@ ARG HOROVOD_WITH_TENSORFLOW=1
...
-ARG HOROVOD_WITHOUT_MPI=1
+ARG HOROVOD_WITHOUT_MPI=0
+ARG HOROVOD_WITH_MPI=1
 ARG HOROVOD_GPU=ROCM
-ARG HOROVOD_WITHOUT_MPI=1
-ENV LD_LIBRARY_PATH=/opt/rocm/lib:/opt/rocm/hip/lib
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH;/opt/rocm/lib:/opt/rocm/hip/lib
+
+#TEST ARGs
+ARG HOROVOD_GPU_ALLREDUCE=NCCL
```
MikhailKardash, line 158: I get an error during `build-pytorch13-tf210-rocm56`:

`Cannot specify both HOROVOD_GPU_OPERATIONS and HOROVOD_GPU_ALLREDUCE options. Try unsetting one of these variables and reinstalling`

Horovod docs state that `HOROVOD_GPU_OPERATIONS` already includes ALLREDUCE and BROADCAST, so it's redundant setting these args:  https://horovod.readthedocs.io/en/stable/install_include.html

will-HPE:
Ilia,

Not that I'm aware of. I suspect we could reorganize the build process so we could build a base image that covers most of the static portions we add and then we we just change the final bits.

Is this what you are getting at?

ioga:
> Not that I'm aware of. I suspect we could reorganize the build process so we could build a base image that covers most of the static portions we add and then we we just change the final bits.
>Is this what you are getting at?

For me, intuitively, ROCm, MPI, libfabric, and infiniband image is a foundational thing and it's not something that should be specific to determined or HPE products. I hoped AMD (or whoever) would already produce this image, and we don't have to take upon this burden...

Dockerfile-default-rocm
will-HPE: removed.

Dockerfile-default-rocm
will-HPE: Commented out lines have been removed.

Dockerfile-default-rocm
will-HPE: A newer version was needed but we're open to a pinned version.

Dockerfile-default-rocm
will-HPE: debugs removed

dockerfile_scripts/scrape_libs.sh
will-HPE: /det_libfabric is needed to prevent multiple libfabrics from being found and breaking the scrape.

dockerfile_scripts/scrape_libs.sh
will-HPE: moved to /var/tmp

Dockerfile-default-rocm
will-HPE: This was changed to:

Dockerfile-default-rocm
will-HPE: fixed

Dockerfile-default-rocm
will-HPE: this has been cleaned up so the correct var is used.

dockerfile_scripts/scrape_libs.sh
```diff
@@ -1,12 +1,20 @@
...
    if [ -d "$host_dir" ]; then
-       libfabric=`find /host -name libfabric.so 2>/dev/null`
+       # to set these paths based on where we find the libs that
+       # libfabric is dependent upon.
+       export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$host_dir/usr/lib64
+       export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$host_dir/usr/local/lib64
+   fi # end if /det_host exists
+   host_dir="/det_libfabric"
+   if [ -d "$host_dir" ]; then
+       libfabric=`find $host_dir -name libfabric.so 2>/dev/null`
```
jerryharrow, line 17: Seems odd to search for it, if the intent of /det_libfabric is to provide libfabric.so

If /det_libfabric is present, should it instead error out with a message if libfabric.so is not found?   That would be a more usable UX.   They have expressed their intent by mounting the directory, so it seems if we ignore the failure it is just going to fail in an unexpected way later, or just not provide the performance that they are expecting.

dockerfile_scripts/scrape_libs.sh
```diff
@@ -1,12 +1,20 @@
 #!/bin/bash
 
+WHOAMI=$(whoami)
 # Only scrape host libs if AWS/NCCL/OFI was built for this image
 if [ -d /container/aws/lib ]
 then
    # See if we mounted in host libs in the expected location
-   host_dir="/host"
+   host_dir="/det_host"
    if [ -d "$host_dir" ]; then
```
karlonw, line 9: Looking at lines 12, 13 and 37 it appears that we're adding "/det_libfabric/...." after "/det_host/...." in the $LD_LIBRARY_PATH search order.  I think we'd want to reverse that, so that if the user provides /det_libfrabic to the container, we definitely want to use the libfabric found there, before any possible different versions that might be in /det_host/usr/lib64 or /det_host/usr/local/lib64.  But this may just be a nit-pick.

Dockerfile-default-rocm
```diff
@@ -1,40 +1,123 @@
...
+                unattended-upgrades \
+        && unattended-upgrade \
+        && rm -rf /var/lib/apt/lists/* \
+        && rm /etc/ssh/ssh_host_ecdsa_key \
+        && rm /etc/ssh/ssh_host_ed25519_key \
+        && rm /etc/ssh/ssh_host_rsa_key
+RUN pip install pip install --upgrade pip
 
 COPY dockerfile_scripts /tmp/det_dockerfile_scripts
+ARG WITH_NCCL
```
MikhailKardash, line 49: What uses `WITH_NCCL` ?

Dockerfile-default-rocm
```diff
@@ -1,40 +1,123 @@
...
+# Install debuild util, etc. for later compiling GDRcopy libraries
+
+# Install GDRcopy so OMPI, NCCL, AWS plugin, etc, can see it
+
+ENV PATH="/opt/conda/envs/py_3.8/bin:${PATH}"
```
MikhailKardash, line 54: This is tied to the specific rocm image version which may or may not have `py_3.8`. Can we parametrize this somehow? 
FWIW, latest image in `rocm` dockerhub has this a step already, so this may not be necessary soon.

Dockerfile-default-rocm
```diff
@@ -1,40 +1,123 @@
...
+
+ARG CONDA="${PATH}"
+
+ENV PYTHONUNBUFFERED=1 PYTHONFAULTHANDLER=1 PYTHONHASHSEED=0
+ARG PYTHON_VERSION
```
MikhailKardash, line 59: What is `PYTHON_VERSION` being used for? Wouldn't this be inferred from the rocm image? Their own arg: 
`ANACONDA_PYTHON_VERSION`

Dockerfile-default-rocm
```diff
@@ -1,40 +1,123 @@
...
+ENV PATH=${PATH:-$CONDA:${WITH_MPI:+$UCX_PATH_DIR:$OMPI_PATH_DIR}}
+
+# Enable running OMPI as root
+ENV OMPI_ALLOW_RUN_AS_ROOT ${WITH_MPI:+1}
+ENV OMPI_ALLOW_RUN_AS_ROOT_CONFIRM ${WITH_MPI:+1}
+
+
+
+RUN pip install cloudpickle
+RUN pip install determined==0.22.1 && pip uninstall -y determined
```
MikhailKardash, line 117: Let's not version pin the `determined` library

Dockerfile-default-rocm
```diff
@@ -48,7 +131,9 @@ RUN /tmp/det_dockerfile_scripts/install_google_cloud_sdk.sh
 # google-api-python-client -> google-api-core -> googleapis-common-protos -> protobuf
 # Horovod cannot build with protobuf > 3.20.x
 # latest google-api-python-client requires protobuf >= 3.20.1
-RUN pip install protobuf==3.20.1
+RUN pip install numpy==1.23.1
```
MikhailKardash, line 134: What is `numpy` required for? It's not a direct dependency of `determined`, right?

Dockerfile-default-rocm
```diff
@@ -63,13 +148,63 @@ ARG HOROVOD_WITH_TENSORFLOW=1
...
+ENV HOROVOD_NCCL_LINK=${WITH_OFI:+SHARED}
+ENV LD_LIBRARY_PATH=${WITH_OFI:+$NCCL_LIB_DIR:}$LD_LIBRARY_PATH
+
 RUN if [ "$HOROVOD_PIP" != "0" ]; then pip install "${HOROVOD_PIP}" ; fi
 
-RUN rm -r /tmp/*
+RUN pip install -r /tmp/det_dockerfile_scripts/additional-requirements.txt
+
+RUN pip uninstall -y tb-nightly tensorboardX
+RUN pip install tensorboard==2.10.1
```
MikhailKardash, line 174: let's put these extra `pip install` things into `additional-requirements.txt`

Dockerfile-default-rocm
```diff
@@ -63,13 +148,63 @@ ARG HOROVOD_WITH_TENSORFLOW=1
...
+
+RUN pip install pynvml
```
MikhailKardash, line 176: see `additional-requirements.txt`

Dockerfile-default-rocm
```diff
@@ -63,13 +148,63 @@ ARG HOROVOD_WITH_TENSORFLOW=1
...
+ARG WITH_AWS_TRACE
+ARG INTERNAL_AWS_DS
+ARG INTERNAL_AWS_PATH
+ARG ROCM_DIR=/opt/rocm
+ENV ROCM_DIR $ROCM_DIR
+RUN if [ "$WITH_OFI" = "1" ]; then /tmp/det_dockerfile_scripts/build_aws_rocm.sh "$WITH_OFI" "$WITH_AWS_TRACE" "$WITH_MPICH"; fi
+ENV LD_LIBRARY_PATH=${WITH_OFI:+$AWS_PLUGIN_INSTALL_DIR:}$LD_LIBRARY_PATH
+RUN ldconfig
+
+RUN pip install tokenizers==0.13.0 huggingface-hub==0.16.4
```
MikhailKardash, line 190: see `additional-requirements.txt`

will-HPE: moved to `additional-requirements-rocm.txt`

Dockerfile-default-rocm
will-HPE: moved to `additional-requirements-rocm.txt`

Dockerfile-default-rocm
will-HPE: Pytorch among other things.

Dockerfile-default-rocm
will-HPE: fixed

Dockerfile-default-rocm
will-HPE: This seems like it is no longer used. I think it was used to try parametrized the Python Version but it didn't yield what was hoped.

Dockerfile-default-rocm
will-HPE: Leaving as is, hopefully isn't needed in the future.

Dockerfile-default-rocm
will-HPE: removed

Dockerfile-default-rocm
```diff
@@ -1,40 +1,123 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 
+RUN apt remove -y openmpi ucx
+#Let's remove existing /opt/ompi; and, link to our version.
```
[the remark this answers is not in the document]
will-HPE: Yes, in the sense that it doesn't have everything we need compiled in.

dockerfile_scripts/scrape_libs.sh
will-HPE: Let's change this in another PR.

dockerfile_scripts/scrape_libs.sh
will-HPE: We are going to address this in a new PR.

dockerfile_scripts/scrape_libs.sh
will-HPE: Rereading this. The /det_libfabric libs are prepended and /det_host /host libs are appended. I think that's what you are seeing no?

dockerfile_scripts/scrape_libs.sh
will-HPE: I've added an error message to explain that `libfabric.so` isn't found within the directory.

PR #227, opened 2023-10-18
build: allow OFI/AWS build of deepspeed [PAD-1]

karlonw:
## Checklist


VERSION
```diff
@@ -1 +1 @@
-0.26.4
+0.26.5
```
karlonw, line 1: Wasn't sure if just adding optional build arguments to a target required a version bump, but bumped anyway. I can revert this change if it is not necessary.

PR #233, opened 2023-11-08
feat: add `ipywidgets` package.

ioga:
## Description

`ipywidgets` provides a better UI for `tqdm` progress bar. it was mentioned a few times over the years on slack, but somehow it has never been transformed into a PR.

## Checklist


VERSION
```diff
@@ -1 +1 @@
-0.26.4
+0.26.5
```
azhou-determined, line 1: next release should be `0.26.4` i think?

azhou-determined (approved):

PR #237, opened 2023-11-21
ci: Pre commit check

MikhailKardash:
## Description

Adding a pre-commit check for the VERSION file to make sure that PRs don't land and accidentally update a released environment. 

## Checklist


.github/workflows/check_version.yml
```diff
@@ -0,0 +1,53 @@
...
+  compare-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Get Determined OSS Version
+        run: |
+          # read determined OSS version file and get the current dev version
+          export OSS_VERSION="$(curl ${{ env.OSS_VERSION_FILE }} | grep -oE '^[0-9.]*')"
+          echo "OSS_VERSION=$OSS_VERSION" >> ${{ github.env }}
```
dannysauer, line 28: Probably better to use the github API since the version file contents are also the name of the latest release and the content of the tag for the latest release.  I doubt that the version file will ever move or that the contents will ever change, but the release naming convention is even less likely to change. :)

Likely the easiest thing here would be to use a third-party action and hide all the sausage-factory
```suggestion
      - name: Get Determined OSS Version
        id: oss
        uses: pozetroninc/github-action-get-latest-release@0.7.0
        with:
          owner: determined-ai
          repo: determined
          excludes: prerelease, draft
```

Then later, instead of using $OSS_VERSION, you'd use `${{ steps.oss.outputs.release }}

Slightly less preferable would be to use `gh api /repos/determined-ai/determined/releases/latest -q .name` instead of the curl to a blob, and store that output as OSS_VERSION.  I don't recall if you need a token for that or not - probably not, since it's a public repo.  If it fails, you just set a global variable to use our repo-wide CI token.

scripts/compare-versions.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
dannysauer, line 1: When wholesale lifting code from StackOverflow, it's probably ideal to give credit. ;)

```suggestion
#!/bin/bash
# based on https://stackoverflow.com/a/4025065/65589
```

dannysauer (approved):
This works as-is, so I'm approving.  There are a couple of comments, but they're "use at your discretion".  However, I wonder if this is the right way to keep the version in-line with OSS?  What happens when this repo has a few updates, but then OSS also goes up a patch level?  What if OSS has several releases before the next update to this repo?

I guess discussing the versioning scheme is not really in scope for this PR.  But I'm inclined to suggest that there's some merit in just tagging this repo's latest release commit every time OSS releases, and then tracking environments versions as a separate thing.  And that sort of brings this comment back in-scope, because it means this action would need to work slightly differently. :)

MikhailKardash:
Closing this. @rb-determined-ai and I agreed that the better direction is to never publish versioned images out of this repo. I think that echoes @dannysauer 's sentiment as well.

PR #238, opened 2023-12-18
chore: Image name change

MikhailKardash:
## Description

Change our image names to be more discoverable. Add a matrix for important package versions.

## Checklist

- [] Bump VERSION to make the pushed images are tagged with the right version.


version_matrix.yaml
```diff
@@ -0,0 +1,57 @@
+git_hash: <git_hash_here>
```
rb-determined-ai, line 1: I don't think I understand what this file is for, if it's got a git_hash embedded into it?

rb-determined-ai (commented):
Overall, looks good to me so far.

To make sure I'm properly understanding the overall plan, we are going to:
a. improve our image naming
b. stop using `determinedai/environments:...` and use `determinedai/the-image-name:...` instead
c. only publish hash images from this repo
d. let the determined repo retag hashes into released versions 
e. switch to NGC+ images

Is that accurate?  It looks like this commit contains a. but not b., d., or e.

What is the overall roll-out plan for how we will deliver this feature to end-users with the fewest number of disruptions?

PR #247, opened 2024-03-04
PAD-127

soohoonchoi:
## Description
Adding a support to build target tf2-gpu using "WITH_OFI" option to support SS11 with GPU nodes.


karlonw (approved):
I don't know enough about the circle-ci process to adequately comment on what will happen, but I'm concerned about two different things:
 1) I'm not sure that the Makefile and build process work identically for the case where "WITH_OFI=0" and the case where "WITH_OFI" is not even set at all.
  2) By adding the with-ofi parameter, will that change existing docker image tags that currently get published without "-ofi-" in their name?  I know we want to create 1 or 2 new images that support ofi, but I dont know if this change will affect many other image names.
 
I'm going to approve it so you can do further testing, but please don't merge it on my say so :-)

PR #248, opened 2024-03-05
PAD-101: Updated build-gpu-ngc-gh-deepspeed to work with llama2 tests

soohoonchoi:
## Description
Update build-gpu-ngc-gh-deepspped for llama2 test.
Added WITH_NCCL support and revert AWS version to 1.4.0.

### DIFF
--- a/dockerfile_scripts/build_aws.sh
+++ b/dockerfile_scripts/build_aws.sh
@@ -17,8 +17,8 @@ if [ "$OFI" = "1" ]; then
                 tcsh
 
   # Install AWS_OFI_NCCL
-  AWS_VER=v1.6.0
-  AWS_VER_NUM=1.6.0
+  AWS_VER=v1.4.0
+  AWS_VER_NUM=1.4.0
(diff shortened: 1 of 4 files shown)


dockerfile_scripts/build_aws.sh
```diff
@@ -17,8 +17,8 @@ if [ "$OFI" = "1" ]; then
                 tcsh
 
   # Install AWS_OFI_NCCL
-  AWS_VER=v1.6.0
-  AWS_VER_NUM=1.6.0
+  AWS_VER=v1.4.0
```
karlonw, line 20: is this correct? going backwards to a previous version of AWS?

otherwise, looks good to me

karlonw (approved):

dockerfile_scripts/build_aws.sh
soohoonchoi: yeah, it seems to work better with the 1.4.0 version.

cdrickett (approved):
Looks good

PR #251, opened 2024-03-13
Add support to build the tf2-gpu image for Libfabric(OFI)

soohoonchoi:
## Description
Add support to build the tf2-gpu image for Libfabric(OFI), which incorporates the AWS libfabric plug-in for NCCL to use on Slingshot 11(SS11) networks.

## Test
Tested the built OFI image by converting to singularity image and tested on the GPU system running SS11: detai-cuda-11.3-pytorch-1.12-tf-2.11-gpu-mpi-ofi-af764fc.sif


MikhailKardash:
Why not just bake OFI into our MPI images, then we don't have to maintain both.
Are they for different use cases? @vishnu2kmohan

cdrickett:
> Why not just bake OFI into our MPI images, then we don't have to maintain both. Are they for different use cases? @vishnu2kmohan
> 
> Going forward, we want a global `-hpc` tag anyway, right?

On IB systems you’ll get much better rdma performance using ucx and not ofi. Historically there are also more bugs when using ofi over verbs on IB than ucx over verbs.

cdrickett:
> > Why not just bake OFI into our MPI images, then we don't have to maintain both. Are they for different use cases? @vishnu2kmohan
> > Going forward, we want a global `-hpc` tag anyway, right?
> 
> On IB systems you’ll get much better rdma performance using ucx and not ofi. Historically there are also more bugs when using ofi over verbs on IB than ucx over verbs.

I suppose if you want only one hpc image how do we distinguish between IB and SS? Or do you not refer to IB as hpc?

karlonw (approved):
Looks good to me

cdrickett (approved):

MikhailKardash:
> > > Why not just bake OFI into our MPI images, then we don't have to maintain both. Are they for different use cases? @vishnu2kmohan
> > > Going forward, we want a global `-hpc` tag anyway, right?
> > 
> > 
> > On IB systems you’ll get much better rdma performance using ucx and not ofi. Historically there are also more bugs when using ofi over verbs on IB than ucx over verbs.
> 
> I suppose if you want only one hpc image how do we distinguish between IB and SS? Or do you not refer to IB as hpc?

Is there a way for users to disable OFI in images?

MikhailKardash:
Also, please update `VERSION` file (to `0.29.2`)

soohoonchoi:
> > > > Why not just bake OFI into our MPI images, then we don't have to maintain both. Are they for different use cases? @vishnu2kmohan
> > > > Going forward, we want a global `-hpc` tag anyway, right?
> > > 
> > > 
> > > On IB systems you’ll get much better rdma performance using ucx and not ofi. Historically there are also more bugs when using ofi over verbs on IB than ucx over verbs.
> > 
> > 
> > I suppose if you want only one hpc image how do we distinguish between IB and SS? Or do you not refer to IB as hpc?
> 
> Is there a way for users to disable OFI in images?

If you are using the OFI image, there is no way to not use OFI with MPI, though for NCCL, you could avoid using OFI by changing the LD_LIBRARY_PATH to not include /container/aws/lib.

MikhailKardash (approved):
approved, but please update VERSION to `0.30.2` before merging.

ioga:
prob 0.31.1 now

PR #254, opened 2024-04-08
Pad 134

will-HPE:
## Tensorboard and GPU kernel build fixes (PAD-91 and PAD-134)
- fix: bug fix

The latest Tensorboard version brought down has issues so we set the required version to one known to work.

## Checklist

- [✓ ] Bump VERSION to make the pushed images are tagged with the right version.


karlonw (approved):
looks good to me

dockerfile_scripts/additional-requirements-rocm.txt
```diff
@@ -10,7 +10,7 @@ Pillow>=8.3.2,<=9.5.0
 analytics-python
 nvidia-ml-py
 protobuf<=3.20.3
-tensorboard==2.10.1
+tensorboard==1.15
```
ioga, line 13: hold up, this can't be right, this is a tensorboard version for tensorflow v1 from 2019.
can we have a version that is more modern?

VERSION
```diff
@@ -1 +1 @@
-0.30.1
+0.31.1
```
ioga, line 1: we've just released 0.30.0; if it'll land (and get into bumpenvs) by EOD today, you can keep it at 0.30.1. otherwise it'll probably be 0.30.2

ioga (changes requested):

dockerfile_scripts/additional-requirements-rocm.txt
will-HPE: We can test a different version but 2.10.1 fails on end-to-end testing and was imortalized in jira PAD-91:

ioga: I have a strong opinion we cannot pin 1.15 here because it's 5 years old and likely to conflict with other dependencies and have CVEs.
2.10.1 is also technically above 1.15...

alexjdetermined (approved):

PR #269, opened 2024-06-24
fix: update nvidia driver version

jgongd:
## Description
MD-413

Fabric manager version and Nvidia driver version mismatch is causing issues.

## Checklist

### DIFF
--- a/Dockerfile-default-rocm
+++ b/Dockerfile-default-rocm
@@ -127,7 +127,10 @@ RUN /tmp/det_dockerfile_scripts/install_google_cloud_sdk.sh
 # latest google-api-python-client requires protobuf >= 3.20.1
 
 ARG TENSORFLOW_PIP
-RUN if [ "$TENSORFLOW_PIP" ]; then pip install --no-cache-dir $TENSORFLOW_PIP; fi
+# Fix flake. The sha256sum of tensorboard_plugin_wit-1.8.1-py3-none-any.whl sometimes
+# doesn't match the one in cache.
+RUN rm -rf ~/.cache/pip
+RUN if [ "$TENSORFLOW_PIP" ]; then pip install $TENSORFLOW_PIP; fi
 
 ARG TORCH_TB_PROFILER_PIP
 RUN if [ "$TORCH_TB_PROFILER_PIP" ]; then pip install $TORCH_TB_PROFILER_PIP; fi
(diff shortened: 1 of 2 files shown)


Dockerfile-default-rocm
```diff
@@ -127,7 +127,7 @@ RUN /tmp/det_dockerfile_scripts/install_google_cloud_sdk.sh
 # latest google-api-python-client requires protobuf >= 3.20.1
 
 ARG TENSORFLOW_PIP
-RUN if [ "$TENSORFLOW_PIP" ]; then pip install $TENSORFLOW_PIP; fi
+RUN if [ "$TENSORFLOW_PIP" ]; then pip install --no-cache-dir $TENSORFLOW_PIP; fi
```
jgongd, line 130: Disable cache to fix flake. Sample CI error:
```
Downloading tensorboard_plugin_wit-1.8.1-py3-none-any.whl (781 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 781.3/781.3 kB 62.8 MB/s eta 0:00:00
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    unknown package:
        Expected sha256 6947733acb94d0596d5b5b694b75065ec5eec7e11b569fea52e2b525ab63e41e
             Got        9013e72119b25d9476ee8185494c4b4de34ebbc120ef711b76f64ef3631a4b3b
```

ioga: does it have to be merged? e.g. perhaps that was a temporary issue? it'd be weird if only one `pip install` in the entire dockerfile has `--no-cache-dir`, and everything else does not.

jgongd: Yeah, I feel it's weird that only one place has `--no-cache-dir`. I have been getting this error from `tensor board_plugin_wit` since 5 days ago https://app.circleci.com/pipelines/github/determined-ai/environments/1099/workflows/cd38934c-e00d-4172-ab7b-ff51e0f3ed70/jobs/13256 and it was still happening yesterday. I thought it might be worthwhile to add it so we can avoid having to manually restart the failed job in CI.

ioga: it's possible that the upstream image somehow already has a pip cache, and it's corrupted in some way. can you please check if doing `rm ~/.cache/pip -rf` early on helps with this? I think it'd be a cleaner way to get rid of the bad cache.

jgongd: Thanks! That helps.

MikhailKardash (approved):
LGTM

PR #276, opened 2024-08-07
test: Add docker image test

gt2345:
## Description
Add test for docker image

## Checklist


Makefile
```diff
@@ -380,3 +380,7 @@ publish-cloud-images:
 		&& packer build $(PACKER_FLAGS) -machine-readable -var "image_suffix=$(SHORT_GIT_HASH)" environments-packer.json \
 		| tee $(ARTIFACTS_DIR)/packer-log
 
+.PHONY: test
+test:
+	docker run --rm -v `pwd`/tests:/workspace/tests -it $(DOCKERHUB_REGISTRY)/$(NGC_PYTORCH_REPO):$(SHORT_GIT_HASH) pytest -m pytorch tests
```
JComins000, line 385: this might fail if I call the make command from a different directory, right?

JComins000 (approved):
ci looks fine

Dockerfile-base-cpu
```diff
@@ -59,6 +59,7 @@ ENV OMPI_ALLOW_RUN_AS_ROOT_CONFIRM ${WITH_MPI:+1}
 # installed.
 RUN pip install determined && pip uninstall -y determined
 
+RUN pip install -r /tmp/det_dockerfile_scripts/requirements.txt
```
MikhailKardash, line 62: We don't ever publish `base` images directly. They're created as a base for our `default` images. This requirement doesn't have to be in both Dockerfiles.

Dockerfile-ngc-hpc
```diff
@@ -3,6 +3,7 @@ FROM ${BASE_IMAGE}
 
 # Copy various shell scripts that group dependencies for install
 COPY dockerfile_scripts /tmp/det_dockerfile_scripts
+RUN pip install -r /tmp/det_dockerfile_scripts/requirements.txt
```
MikhailKardash, line 6: This doesn't have to be here either, as these images are always built from either `Dockerfile-pytorch-ngc` or `Dockerfile-tensorflow-ngc`

Dockerfile-pytorch-ngc
```diff
@@ -20,6 +20,7 @@ RUN /tmp/det_dockerfile_scripts/install_libnss_determined.sh
 # installed.
 RUN python -m pip install determined && python -m pip uninstall -y determined
 
+RUN pip install -r /tmp/det_dockerfile_scripts/requirements.txt
```
MikhailKardash, line 23: This is installed on line 26.

MikhailKardash (commented):
We don't need to install `pytest` as a requirement in the images. Let's `pip install pytest` after building. This can be done as a one-liner after `docker run /bin/bash -c "pip install pytest && pytest ...."` or as a separate shell script.

MikhailKardash:
Regarding the python packages being tested:
1. `gpt-neox` images should also test for `pytorch`
2. pytorch-tensorflow (`default-cpu` and `default-cuda`) images need both `pytorch` and `tensorflow` tests
3. `ngc-pytorch` images contain `deepspeed`, so let's test for that.

Makefile
```diff
@@ -380,3 +392,7 @@ publish-cloud-images:
 		&& packer build $(PACKER_FLAGS) -machine-readable -var "image_suffix=$(SHORT_GIT_HASH)" environments-packer.json \
 		| tee $(ARTIFACTS_DIR)/packer-log
 
+.PHONY: test
+test:
+	docker run --rm -v `pwd`/tests:/workspace/tests -it $(DOCKERHUB_REGISTRY)/$(NGC_PYTORCH_REPO):$(SHORT_GIT_HASH) /bin/bash -c "pip install pytest && pytest -m pytorch /workspace/tests"
+	docker run --rm -v `pwd`/tests:/workspace/tests -it $(DOCKERHUB_REGISTRY)/$(NGC_TF_REPO):$(SHORT_GIT_HASH) /bin/bash -c "pip install pytest && pytest -m pytorch /workspace/tests"
```
MikhailKardash, line 398: Missing end newlines in all 3 changed files

Makefile
```diff
@@ -380,3 +392,7 @@ publish-cloud-images:
 		&& packer build $(PACKER_FLAGS) -machine-readable -var "image_suffix=$(SHORT_GIT_HASH)" environments-packer.json \
 		| tee $(ARTIFACTS_DIR)/packer-log
 
+.PHONY: test
+test:
+	docker run --rm -v `pwd`/tests:/workspace/tests -it $(DOCKERHUB_REGISTRY)/$(NGC_PYTORCH_REPO):$(SHORT_GIT_HASH) /bin/bash -c "pip install pytest && pytest -m pytorch /workspace/tests"
```
MikhailKardash, line 397: Don't need this test workflow since it isn't used anywhere.

Makefile
```diff
@@ -325,6 +336,7 @@ build-pytorch-cuda: build-cuda-118-base
 		-t $(DOCKERHUB_REGISTRY)/$(CUDA_PYTORCH_ENVIRONMENT_NAME):$(SHORT_GIT_HASH) \
 		-t $(NGC_REGISTRY)/$(CUDA_PYTORCH_ENVIRONMENT_NAME):$(SHORT_GIT_HASH) \
 		.
+	docker run --rm -v `pwd`/tests:/workspace/tests -it $(DOCKERHUB_REGISTRY)/$(CUDA_PYTORCH_ENVIRONMENT_NAME):$(SHORT_GIT_HASH) /bin/bash -c "pip install pytest && pytest -m pytorch /workspace/tests"
 
 .PHONY: publish-tensorflow-cpu
 publish-tensorflow-cpu:
```
MikhailKardash, line 342: Should have a test here as well

Makefile
```diff
@@ -309,6 +319,7 @@ build-pytorch-cpu: build-cpu-py-310-base
 		$(CPU_PYTORCH_TAGS) \
 		--push \
 		.
+	docker run --rm -v `pwd`/tests:/workspace/tests -it $(DOCKERHUB_REGISTRY)/$(CPU_PYTORCH_ENVIRONMENT_NAME):$(SHORT_GIT_HASH) /bin/bash -c "pip install pytest && pytest -m pytorch /workspace/tests"
```
MikhailKardash, line 322: These multi-arch images seem to be pulling with an error `Requested platforms linux/amd64 do not match result platforms linux/amd64,linux/arm64`
You might have to add the `--platform` arg to the run command for this, and `tensorflow-cpu`.

Makefile
gt2345: You mean a test for `publish-tensorflow-cpu`?

Makefile
gt2345: Is this similar to how we use `--platform` when build? I don't see the error though

MikhailKardash (approved):
LGTM

